PR #13, opened 2016-07-11
Rename adler32 to adler32_zlib

jibsen:
The zlib library has an `adler32()` function. This renames the function in libdeflate to avoid conflicts if both libraries are linked. I used `adler32_zlib()` to equal the naming on `crc32_gzip()`.

### DIFF
--- a/lib/adler32.c
+++ b/lib/adler32.c
@@ -46,7 +46,7 @@
 #define UNROLL_FACTOR	4
 
 u32
-adler32(const void *buffer, size_t size)
+adler32_zlib(const void *buffer, size_t size)
 {
 	u32 s1 = 1;
 	u32 s2 = 0;
--- a/lib/adler32.h
+++ b/lib/adler32.h
@@ -7,6 +7,6 @@
 
 #include "common_defs.h"
 
-extern u32 adler32(const void *buffer, size_t size);
+extern u32 adler32_zlib(const void *buffer, size_t size);
 
 #endif /* _LIB_ADLER32_H */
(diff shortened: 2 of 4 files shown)


ebiggers:
This only affects static linking, correct?  `adler32()` is not yet part of libdeflate's API and is not exported from the shared library.

The reason for the `crc32_gzip()` name was that there are multiple types of CRC-32s, and the one that is implemented is specifically the one used in the gzip format.  This is not the case with Adler-32, of which there is only one version.

I would like there to be a more general solution but I'm not sure there is, other than prefixing all global symbols (even ones not part of the API) with `libdeflate_`.

jibsen:
Yes, it only affects static linking.

I tried compiling the benchmark example program with static libdeflate and zlib, and depending on the order of the libraries on the command line when linking, either zlib or libdeflate crashes the program (due to the two adler32 implementations being incompatible). I guess MSVC picks whatever version of adler32 it finds first.

Adding a prefix to all globals would solve this as well (and is probably a good idea in general).

Either way, I am adding libdeflate to pngwolf-zopfli, but also need zlib for the streaming interface, and would prefer to be able to statically link them both.

ebiggers:
I think I'm going to make an internal library header which redefines the non-API global symbols to be prefixed with "_libdeflate".  I still don't like the solution, but at least it can be made to apply to all relevant symbols, not just `adler32()`, and the actual code can effectively keep using the current, less ugly names.

ebiggers:
I committed my solution to master in bcc741cc7b43d1658097f7d279e86aadce13e626.  So static linking with both libdeflate and zlib should work now.

Thanks for the suggestions and for testing libdeflate!

jibsen:
Thank you for the quick resolution!

One note; identifiers that start with an underscore are reserved in C.

PR #16, opened 2016-09-03
Update handling of exports

jibsen:
When compiling libdeflate on Windows, the public functions are declared dllexport. This results in executables that are linked with libdeflate also exporting these functions.

This is an attempt to fix this, and gather the export handling in the main header. I don't know if you prefer some other way to fix this though.

### DIFF
--- a/Makefile.msc
+++ b/Makefile.msc
@@ -5,6 +5,9 @@
 #	nmake /f Makefile.msc
 #
 
+.SUFFIXES:
+.SUFFIXES: .obj .dllobj .c
+
 CC = cl
 LD = link
 AR = lib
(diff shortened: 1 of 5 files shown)


ebiggers:
Thanks for the suggestion.  This actually is a broader problem as it affects static libraries built for UNIX or with MinGW, not just with MSVC.  The proposed change also breaks exporting symbols from the shared library when not compiling with MSVC.

I've pushed a reworked change up to the branch "static_library_fix".  It should handle all supported compilers and platforms correctly.  If you want to test it with your setup, that would be great.  Thanks!

jibsen:
Your branch fixes the issue I had as well, thanks.

ebiggers:
I commited my version of the change to master.  Thanks!

PR #18, opened 2016-09-23
gzip.c did not compile under MacOSX:fixed this with preprocessor stuf…

maarten-k:
…f to keep things portable

### DIFF
--- a/programs/gzip.c
+++ b/programs/gzip.c
@@ -37,6 +37,16 @@
 #  include <utime.h>
 #endif
 
+
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


ebiggers:
Thanks for the suggestion!  I actually decided to go with a more general solution which is to detect which fields are actually present, rather than assuming that OS X in particular uses the nonstandard names.  It should also handle other operating systems correctly, e.g. older BSDs which it sounds like use the nonstandard names too.

I commited my version to master.  Let me know if it doesn't actually work on OS X for some reason (I don't have access to an OS X system right now).

PR #29, opened 2017-06-26
ARM Neon CRC support

JunHe77:
Change-Id: If24a5a7535ee29e7535264a3adbfc42ab26f8e59
Signed-off-by: Jun He <jun.he@linaro.org>

### DIFF
--- a/lib/crc32_arm.h
+++ b/lib/crc32_arm.h
@@ -1,0 +1,162 @@
+/*
+ * crc32_arm.h
+ *
+ * Copyright 2017 Jun He <jun.he@linaro.org>
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
+ * restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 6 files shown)


JunHe77:
Tested on aarch64 using tools/checksum_benchmarks.sh:
Method                              Speed (MB/s)
------                              ------------
CRC-32 (libdeflate, NEON-MULL)      6593
CRC-32 (libdeflate, generic)        1044
CRC-32 (zlib)                       660

Adler-32 (libdeflate, SIMD(asimd))  6391
Adler-32 (zlib)                     1384
Adler-32 (libdeflate, generic)      1272

ebiggers:
Thanks for doing this!  However, when I compile libdeflate for aarch64, the new code is not compiled unless I explicitly specify `CFLAGS=-marmv8-a+crypto`.  Shouldn't it be possible to always compile it, then select it at runtime based on the CPU features?  That's what's done for the x86 PCLMUL implementation.

lib/crc32_arm.h
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/*
+ *
+ * Note: on unaligned ends of the buffer, we fall back to crc32_slice1() instead
+ * of crc32_slice8() because only a few bytes need to be processed, so a smaller
+ * table is preferable.
+ */
+static u32 ATTRIBUTES
+FUNCNAME(u32 remainder, const u8 *buffer, size_t nbytes)
```
ebiggers: No need for the FUNCNAME abstraction if this header isn't going to be include multiple times; it should just be called `crc32_pmull`.  Similarly for FUNCNAME_ALIGNED and ATTRIBUTES.

JunHe77:
Updated.

ebiggers:
Sorry for the slow review, I've been busy.  There are still some problems with the current proposal.  First, it doesn't compile on x86 anymore because it always tries to include `<asm/hwcap.h>`, which doesn't exist.  Also it appears that `<asm/hwcap.h>` is Linux-specific, so it will also break compiling libdeflate for non-Linux ARM.  Furthermore I couldn't even get it to compile for Android (which is both ARM and Linux); for arm32 it couldn't find `<sys/auxv.h>` and for arm64 it failed too because `HWCAP_ATOMICS` is undeclared.  So unless these definitions are in a standard place for gcc and clang (not depending on the specific OS or C library) then they will need to be defined in the libdeflate source instead.

Second, forcing a particular `-march` and/or `-mfpu` setting is a bad idea because the user should be able to choose the target architecture themselves.  It's especially problematic if the user wants the code to run on a CPU that doesn't have the features forced on by the CFLAGS, since in that case it would crash at runtime.  Please take a close look at how I solved this problem for x86: by default (with no `-march` option given) the compiler will assume that pclmul is unavailable, so `__PCLMUL__` will be undefined, and instead we compile the pclmul-accelerated function using `__attribute__((target("pclmul")))` and only call it at runtime if cpuid says that pclmul is available.  But if the user *does* compile specifically for a CPU that has pclmul (e.g. using `make CFLAGS=-march=native` on a system running a recent Intel processer), then `__PCLMUL__` will be defined so the target function attribute isn't needed, and `crc32_pclmul()` is called unconditionally.  Why not do something similar for ARM?

ebiggers:
Reworked version committed as fc2ea22b4483e.  Thanks!

PR #32, opened 2018-05-12
Set UNALIGNED_ACCESS_IS_FAST on powerpc64

antonblanchard:
All 64bit PowerPC CPUs handle unaligned accesses reasonably fast, so
set UNALIGNED_ACCESS_IS_FAST.

Decompression of the snappy html test case is almost 50% faster on
POWER9 with this patch applied.


ebiggers:
Looks good, thanks!

PR #38, opened 2018-12-25
Makefile: make the installation prefix configurable

orivej:
This will be useful in [Nixpkgs](https://github.com/NixOS/nixpkgs/) since for example the library needs to be installed in `/nix/store/{hash}-libdeflate-{version}/lib/libdeflate.so`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,8 @@ override CFLAGS :=							\
 
 ##############################################################################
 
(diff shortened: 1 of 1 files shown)


orivej:
Updated NEWS.

ebiggers:
Merged, thanks!

PR #43, opened 2019-04-02
Build dylib on MacOS

epruesse:
- Using `ar cr` yields architecture errors with default toolchain, using
  `libtool -static -o` instead seems to work.
- MacOS builds a flat archive, so duplicate names as used for `cpu_features.o` do not work. Fixed by disabling adding of ARM object file(s) to archive. 
- On MacOS, libs end in `dylib`
- On MacOS, SOVERSION comes between basename and extension.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,8 +77,7 @@ else ifeq ($(shell uname),Darwin)
    SHARED_LIB_SUFFIX  := .dylib
    SHARED_LIB          = libdeflate$(SOVERSION)$(SHARED_LIB_SUFFIX)
    SHARED_LIB_LDFLAGS := "-install_name $(SHARED_LIB)"
-   AR                 := libtool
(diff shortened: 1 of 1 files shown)


ebiggers:
Hi, this looks useful, though I don't have a way to test MacOS support myself at the moment so I really need you to do this properly.  I notice this pull request includes 3 commits, but the second and third appear to be fixes for the first, and they leave the `ARFLAGS` and `ARCHS` variables which appear to be unnecessary.  Before I take a closer look, can you please squash your changes together into one coherent commit that doesn't leave any debris?  Please also make sure the changes are described in the commit itself and not just in pull request description.  Thanks!

ebiggers:
Reworked version committed as 1a440df ("Makefile: build dylib on macOS").  Thanks.

PR #45, opened 2019-06-30
AArch64: feature detection on FreeBSD, faster native CRC32 instructions

myfreeweb:
- assume NEON (ASIMD) support on AArch64, since it's mandatory for all processors
- detect features on FreeBSD >= 12 via `ID_AA64ISAR0_EL1`
- add support for native CRC32 instructions for even more performance (results from Marvell Armada 8040 @ 2ghz)

```
cebda3bb        "/boot/kernel/kernel"   14 ms   975 MB/s    # fallback
cebda3bb        "/boot/kernel/kernel"   2 ms    5735 MB/s   # pmull
cebda3bb        "/boot/kernel/kernel"   1 ms    14049 MB/s  # crc32 instr
```

(not that it affects actual compression by much, gzipping the same file is ~204MB/s with fallback, ~247MB/s with pmull, ~249MB/s with crc32 instructions)


lib/arm/cpu_features.c
```diff
@@ -116,4 +118,34 @@ void setup_cpu_features(void)
 	_cpu_features = features | ARM_CPU_FEATURES_KNOWN;
 }
 
+#elif defined(__aarch64__) /* not __linux__ */
+
+volatile u32 _cpu_features = 0;
```
ebiggers: Move this to just below `#if ARM_CPU_FEATURES_ENABLED`, so it's one place for the Linux and non-Linux cases.

lib/arm/cpu_features.c
```diff
@@ -116,4 +118,34 @@ void setup_cpu_features(void)
...
+	/* NEON is a mandatory part of AArch64 */
+	_cpu_features = ARM_CPU_FEATURE_NEON;
+
+	/* On FreeBSD >= 12, Linux >= 4.11 and potentially others, it is possible to use
+	   privileged system registers from userspace to check CPU feature support.
+
+	   For proper support of SoCs where different cores have different capabilities,
+	   the OS has to always report only the features supported by all cores, like FreeBSD does. */
+
+#define bits_shift(x, high, low) ((x >> low) & ((1 << (high - low + 1)) - 1))
```
ebiggers: Call this `bitfield()` instead of `bits_shift()`?

lib/arm/cpu_features.c
```diff
@@ -116,4 +118,34 @@ void setup_cpu_features(void)
...
+
+#if __FreeBSD__ >= 12
+	uint64_t isar0 = 0;
```
ebiggers: For consistency: `u64`, not `uint64_t`

lib/arm/cpu_features.c
```diff
@@ -116,4 +118,34 @@ void setup_cpu_features(void)
...
+	__asm__("mrs %0, ID_AA64ISAR0_EL1" : "=r"(isar0));
+
+	if (bits_shift(isar0, 7, 4) >= 1)
```
ebiggers: Did you check the [AArch64 documentation](https://developer.arm.com/docs/ddi0595/latest/aarch64-system-registers/id_aa64isar0_el1)?  AFAICS, this check needs to be `== 2`, not `>= 1`.

lib/arm/crc32_impl.h
```diff
@@ -3,6 +3,7 @@
  *
  * Copyright 2017 Jun He <jun.he@linaro.org>
  * Copyright 2018 Eric Biggers
+ * Copyright 2019 Greg V
```
ebiggers: Can you please include your full last name?

lib/arm/crc32_impl.h
```diff
@@ -151,12 +152,66 @@ crc32_pmull_aligned(u32 remainder, const uint8x16_t *p, size_t nr_segs)
...
+     /* not yet tested on big endian, probably needs changes to work there */ \
+    (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+#  define FUNCNAME		crc32_instr
+#  define FUNCNAME_ALIGNED	crc32_instr_aligned
+#  define FUNCNAME_SLICE1	crc32_instr_slice1
+#  ifdef __ARM_FEATURE_CRC32
+#    define ATTRIBUTES
+#    define DEFAULT_IMPL	crc32_instr
+#  else
+#    define __ARM_FEATURE_CRC32	1
```
ebiggers: What is this line for?

lib/arm/crc32_impl.h
```diff
@@ -151,12 +152,66 @@ crc32_pmull_aligned(u32 remainder, const uint8x16_t *p, size_t nr_segs)
...
+    (ARM_CPU_FEATURES_ENABLED &&	\
+     COMPILER_SUPPORTS_CRC32_TARGET_INTRINSICS)) && \
+     /* not yet tested on big endian, probably needs changes to work there */ \
+    (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+#  define FUNCNAME		crc32_instr
+#  define FUNCNAME_ALIGNED	crc32_instr_aligned
+#  define FUNCNAME_SLICE1	crc32_instr_slice1
+#  ifdef __ARM_FEATURE_CRC32
+#    define ATTRIBUTES
+#    define DEFAULT_IMPL	crc32_instr
```
ebiggers: `DEFAULT_IMPL` may have already been defined.  How are you trying to order the default implementations?  I.e. if someone compiles with both pmull and crc32 instructions enabled, which implementation do you intend to use?  How does performance compare?  Check `lib/x86/crc32_impl.h` for how this is handled for x86.  E.g. if the crc32 instructions are meant to be first priority, you have to define it first, and then only define the pmull implementation if `!defined(DEFAULT_IMPL)` already.

lib/arm/crc32_impl.h
```diff
@@ -151,12 +152,66 @@ crc32_pmull_aligned(u32 remainder, const uint8x16_t *p, size_t nr_segs)
 #include "../crc32_vec_template.h"
 #endif /* PMULL implementation */
 
+#undef DISPATCH_INSTR
+#if (defined(__ARM_FEATURE_CRC32) ||	\
+    (ARM_CPU_FEATURES_ENABLED &&	\
+     COMPILER_SUPPORTS_CRC32_TARGET_INTRINSICS)) && \
+     /* not yet tested on big endian, probably needs changes to work there */ \
+    (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
```
ebiggers: How about just putting in the `le64_bswap()`s that are needed for it to work on big endian?

lib/arm/crc32_impl.h
```diff
@@ -151,12 +152,66 @@ crc32_pmull_aligned(u32 remainder, const uint8x16_t *p, size_t nr_segs)
 #include "../crc32_vec_template.h"
 #endif /* PMULL implementation */
 
+#undef DISPATCH_INSTR
+#if (defined(__ARM_FEATURE_CRC32) ||	\
+    (ARM_CPU_FEATURES_ENABLED &&	\
```
ebiggers: Can you please fix the indentation to make it clear that these parts are all one clause of the `||`?  See the existing code.

lib/arm/crc32_impl.h
```diff
@@ -151,12 +152,66 @@ crc32_pmull_aligned(u32 remainder, const uint8x16_t *p, size_t nr_segs)
...
+		remainder = __crc32d(remainder, p[i * 4 + 0]);
+		remainder = __crc32d(remainder, p[i * 4 + 1]);
+		remainder = __crc32d(remainder, p[i * 4 + 2]);
+		remainder = __crc32d(remainder, p[i * 4 + 3]);
+	}
+	return remainder;
+}
+#define IMPL_ALIGNMENT		8
+#define IMPL_SEGMENT_SIZE	32
+#include "../crc32_vec_template.h"
```
ebiggers: It's a bit messy using `crc32_vec_template.h` for this because you have to override the SLICE1 function.  Did you consider just coding `crc32_insr()` as a standalone function?  That might be cleaner.

lib/arm/cpu_features.c
```diff
@@ -111,9 +113,43 @@ void setup_cpu_features(void)
...
+		_cpu_features |= ARM_CPU_FEATURE_PMULL;
+	if (bits_shift(isar0, 19, 16) >= 1)
```
ebiggers: [AArch64 documentation](https://developer.arm.com/docs/ddi0595/latest/aarch64-system-registers/id_aa64isar0_el1) says this bitfield contains a value, not individual bits, and only 0 and 1 are defined currently.  So I think this should be `== 1`, not `>= 1`.

ebiggers:
Oh, I missed that the benchmark results were in the pull request description.  Can you move them to the commit message instead so they don't get lost?  Thanks!!

lib/arm/cpu_features.c
myfreeweb: Yes, technically "From Armv8.1, the only permitted value is 0b0001", but all these bitfields work like increasing levels of support (look at AES and SHA2).

lib/arm/crc32_impl.h
myfreeweb: The header only defines `__crc32..` (as equivalent to `__builtin_...crc32..`) when that is defined (i.e. when the target feature is enabled, which we're doing here via an attribute, but that doesn't set these defines as that happens after the preprocessor).

As I understand, `__crc32..` is the public API.

lib/arm/crc32_impl.h
myfreeweb: No. Actually, I don't really need a copyright line..

lib/arm/crc32_impl.h
myfreeweb: ehh, that would be copy-pasting the alignment code from the template..

lib/arm/crc32_impl.h
ebiggers: It works for me even I delete that line (I'm using gcc 8.2.1 on Linux on AArch64).

Normally the CPU feature macros like `__ARM_FEATURE_*`, `__AVX2__`, etc. are meant to be defined by the compiler based on the compilation flags, not by the user.  And intrinsics are available in `target` attribute functions regardless of these, though sometimes not with old compilers.  The way I've handled this in libdeflate elsewhere is just to disable the code if the compiler is too old.  I don't want to go too far into the weeds of trying to make all the optimizations work with old compilers.

I don't see why we shouldn't do the same here.  Can you explain what compiler versions need this, exactly?

myfreeweb: > intrinsics are available in target attribute functions regardless of these

That's not the case with clang.

https://github.com/llvm/llvm-project/blob/18b0c40bc5dbaf25157a838701b2aba8ba8440db/clang/lib/Headers/arm_acle.h#L558

(this is the newest revision)

ebiggers: Well, I guess we can do it then, but please leave a brief comment explaining that it's a workaround for clang.  Can you also check whether there's any clang bug report for this, and if not, file one?  A more minimal example:

```
#include <arm_neon.h>
#include <arm_acle.h>

poly128_t __attribute__((target("crypto")))
crypto_insn(poly64_t a)
{
        return vmull_p64(a, a);
}

uint32_t __attribute__((target("crc")))
crc32_insn(uint32_t crc, uint8_t byte)
{
        return __crc32b(crc, byte);
}
```

With default flags, i.e. no crypto or crc enabled, clang can compile the first function, but not the second function.  gcc can compile both.  (Well, for gcc you have to prepend "+" to "crypto" and "crc".)

ebiggers:
Closing stale pull requests with unaddressed comments. If you still want this merged please update it.

Issue #30, opened 2017-06-26
filename and timestamps storing


cieplutki:
Hey,
is gzip able to save file name or time stams of compressed file?
I looked into options but there seem to be nothing which suggests there is no such option.

ebiggers:
No, storing those fields is not implemented yet.  I think they are rarely needed, actually.  Why do you need them?

cieplutki:
I use it as archiver not only compressor and so I don't have to do this: 
```
$ tar c file.name | gzip - > file.name.tar.gz
(...)
$ tar xf file.name.gz | gunzip
```

Sometimes I have to deal with damaged/recovered archives and filename at that point does make a difference. If you have yet another "anonymous" file that god one knows whats inside - there is a chance it will be "ignored" on purpose where there are tens or hundreds of similar ones recovered from file system crash. In my experience the most frequent fs crashes "produce" files that are "cut off" at the end. Whenever you try to undelete or recover raw files (without filesystem - name/directory structure) they are somewhat incomplete at the end (due to file fragmentation) and then name of file can give you a clue whether it's what you are looking for or not.

PS. Sorry for lengthy explanation, hope it is not off-topic.

Dobatymo:
The name gzip suggests a gzip alternative... The original software stores filename and modtime by default. So I wouldn't say it's a rarely needed feature...

PR #50, opened 2019-08-07
Solution for issue #30

JosePineiro:
Activate -N switch.
Optional store filename in GZ

Create a new function libdeflate_gzip_compress_ex for preserve compatibility. The old function libdeflate_gzip_compress have the same parameters thet the previous version.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+Version 1.3:
+	Activate -n switch in gzip program
+
+	Add file_name parameter to libdeflate_gzip_compress function. This
+	function store filename in gz if file_name parameter in not null.
+	
 Version 1.2:
 	Slight improvements to decompression speed.
 
--- a/lib/gzip_compress.c
+++ b/lib/gzip_compress.c
@@ -34,9 +34,10 @@
 #include "libdeflate.h"
 
 LIBDEFLATEAPI size_t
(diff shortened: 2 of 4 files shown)


NEWS
```diff
@@ -1,3 +1,9 @@
+Version 1.3:
+	Activate -n switch in gzip program
+
+	Add file_name parameter to libdeflate_gzip_compress function. This
```
ebiggers: Why isn't the decompression case implemented?  If this feature is offered at all, I think people would expect that it be supported on both the compression and decompression side.

lib/gzip_compress.c
```diff
@@ -66,7 +70,14 @@ libdeflate_gzip_compress(struct libdeflate_compressor *c,
 		xfl |= GZIP_XFL_SLOWEST_COMRESSION;
 	*out_next++ = xfl;
 	/* OS */
-	*out_next++ = GZIP_OS_UNKNOWN;	/* OS  */
+	*out_next++ = GZIP_OS_UNKNOWN;	
+	/* FILENAME */
```
ebiggers: `/* Original file name (zero terminated) */`, to match the text in the gzip specification and the comment in gzip_decompress.c.

ebiggers:
Closing stale pull requests with unaddressed comments.  If you still want this merged please update it.

PR #63, opened 2020-04-11
Add function libdeflate_zlib_decompress_ex

IzumiRaine:
I need to decompress a zlib-compressed buffer with some additional trailing junk bytes. To handle this, I've added the function `libdeflate_zlib_decompress_ex` with an additional argument `actual_in_nbytes_ret`, analogously to the gzip functions.

Additionally, libdeflate_zlib_decompress now returns successfully in case there are additional trailing bytes in the input buffer after the compressed stream.

### DIFF
--- a/libdeflate.h
+++ b/libdeflate.h
@@ -263,6 +263,19 @@ libdeflate_zlib_decompress(struct libdeflate_decompressor *decompressor,
 			   size_t *actual_out_nbytes_ret);
 
 /*
+ * Like libdeflate_zlib_decompress(), but adds the 'actual_in_nbytes_ret'
+ * argument.  If 'actual_in_nbytes_ret' is not NULL and the decompression
+ * succeeds, then the actual number of input bytes consumed is written to
+ * actual_in_nbytes_ret.
+ */
+LIBDEFLATEEXPORT enum libdeflate_result LIBDEFLATEAPI
+libdeflate_zlib_decompress_ex(struct libdeflate_decompressor *decompressor,
+			      const void *in, size_t in_nbytes,
+			      void *out, size_t out_nbytes_avail,
+			      size_t *actual_in_nbytes_ret,
+			      size_t *actual_out_nbytes_ret);
+
+/*
  * Like libdeflate_deflate_decompress(), but assumes the gzip wrapper format
  * instead of raw DEFLATE.
  *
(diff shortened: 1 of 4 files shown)


libdeflate.h
```diff
@@ -262,6 +262,19 @@ libdeflate_zlib_decompress(struct libdeflate_decompressor *decompressor,
 			   void *out, size_t out_nbytes_avail,
 			   size_t *actual_out_nbytes_ret);
 
+/*
+ * Like libdeflate_zlib_decompress(), but adds the 'actual_in_nbytes_ret'
+ * argument.  If 'actual_in_nbytes_ret' is not NULL and the decompression
+ * succeeds, then the actual number of input bytes consumed is written to
```
ebiggers, line 268: This isn't clear about when the decompression stops.  Can you mention that it decompresses the first zlib-compressed stream (similar to the comment for `libdeflate_gzip_decompress_ex()`)?:

```
/*
 * Like libdeflate_zlib_decompress(), but adds the 'actual_in_nbytes_ret'
 * argument.  If 'actual_in_nbytes_ret' is not NULL and the decompression
 * succeeds (indicating that the first zlib-compressed stream in the input
 * buffer was decompressed), then the actual number of input bytes consumed is
 * written to *actual_in_nbytes_ret.
 */
```

It would also be helpful to update the comment for `libdeflate_zlib_decompress()` to clarify that it stops at the end of the zlib stream:

```
/*
 * Like libdeflate_deflate_decompress(), but assumes the zlib wrapper format
 * instead of raw DEFLATE.
 *
 * Decompression will stop at the end of the zlib stream, even if it is shorter
 * than 'in_nbytes'.  If you need to know exactly where the zlib stream ended,
 * use libdeflate_zlib_decompress_ex().
 */

```

ebiggers:
Also, the CI failure with "FreeBSD:freebsd-11-2-release-amd64" should be fixed if you rebase onto the master branch.

IzumiRaine:
Thanks for your comments. I've added some tests. Are they good or did you have something different in mind?

(As for why I need this: I'm trying to parse an old file format which sometimes adds padding after the compressed data, but unfortunately it doesn't say how much padding there is.)

ebiggers:
Thanks!  There were some compiler warnings in the test (apparently Travis CI isn't running on pull requests; I need to figure out why...), but I ended up rewriting it anyway in  21ccbd2e2f855a18657bc38c7e8af88b3ec9eb0c to make it cover DEFLATE and gzip too.

IzumiRaine:
Great, thanks for merging! (I hadn't noticed any compiler warnings, but maybe that's because I'm on macOS using clang.)

PR #66, opened 2020-06-27
README.md: add macOS subsection

SeekingMeaning:


README.md
```diff
@@ -26,6 +26,14 @@ use this library are also provided:
 
 # Building
 
+## For macOS
```
ebiggers, line 29: Can you please put this after the "For UNIX" section rather than before?

Then replace "below" => "above".

ebiggers:
Squashed and merged manually as https://github.com/ebiggers/libdeflate/commit/675be1ccf8958cc0d76a5f5d234273bd6442250c.  Thanks!

PR #68, opened 2020-07-10
Fixup -D FREESTANDING on MSVC

RReverser:
MSVC doesn't have an option for proper freestanding support, but there's no reason why compilation with this flag should completely fail.

Before this change, the `#define malloc NULL` and `#define free NULL` were followed by `#include <stdlib.h>` and conflicted with real declarations, resulting in weird syntax errors.

Note that this issue didn't occur in the original commit - https://github.com/RReverser/libdeflate/commit/7c7fed4b384dda1e54101229a258694a89bad735 - but, looks like it was modified upon merge from `stddef.h` to `stdlib.h` here https://github.com/ebiggers/libdeflate/commit/e2d1621e421e4a65bfe04d7f2687445612f91654. Judging from the description, this was done for backward compatibility with old MSVC?

Anyway, this PR fixes the issue, while keeping the `stdlib.h` include, by simply reordering custom definitions to appear after the standard ones.

Additionally, it changes conditional block on the recently introduced `__builtin_mem*` functions to use regular `mem*` functions in case -D FREESTANDING is used on MSVC, because the latter doesn't have `__builtin_mem*`s.

In the end, the `-D FREESTANDING` compiles on MSVC without issues again, but the only actual effect it has is changing default allocator functions to `NULL`.

This is still useful for cross-platform compatibility of builds with same flags.

### DIFF
--- a/lib/utils.c
+++ b/lib/utils.c
@@ -25,14 +25,14 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#include "lib_common.h"
+
 #ifdef FREESTANDING
 #  define malloc NULL
 #  define free NULL
 #else
 #  include <stdlib.h>
 #endif
-
-#include "lib_common.h"
 
 #include "libdeflate.h"
 
--- a/common/compiler_msc.h
+++ b/common/compiler_msc.h
@@ -3,7 +3,7 @@
  */
 
 #include <stdint.h>
-#include <stddef.h> /* for _byteswap_*() */
+#include <stdlib.h> /* for _byteswap_*() */
 
 #define LIBEXPORT	__declspec(dllexport)
 
--- a/common/compiler_msc.h
+++ b/common/compiler_msc.h
@@ -3,7 +3,7 @@
  */
 
 #include <stdint.h>
-#include <stdlib.h> /* for _byteswap_*() */
+#include <stddef.h> /* for _byteswap_*() */
 
 #define LIBEXPORT	__declspec(dllexport)
 
--- a/lib/lib_common.h
+++ b/lib/lib_common.h
@@ -36,7 +36,7 @@ void libdeflate_free(void *ptr);
 void *libdeflate_aligned_malloc(size_t alignment, size_t size);
 void libdeflate_aligned_free(void *ptr);
 
-#ifdef FREESTANDING
+#if defined(FREESTANDING) && !defined(_MSC_VER)
 /*
  * With -ffreestanding, <string.h> may be missing, and we must provide
  * implementations of memset(), memcpy(), memmove(), and memcmp().


ebiggers:
I'm confused.  If MSVC doesn't support freestanding builds, why would you expect -DFREESTANDING to work with it?

Also, the second commit in this pull request is just undoing a change made by the first.  Can you squash them together and write a correct commit message?

Also, the first commit message says that your original commit https://github.com/RReverser/libdeflate/commit/7c7fed4b384dda1e54101229a2 didn't have this problem.  But actually compiler_msc.h still included <stdlib.h> in that version too (it's later in the file, not at the top).

RReverser:
> I'm confused. If MSVC doesn't support freestanding builds, why would you expect -DFREESTANDING to work with it?

As I said in the PR: "This is still useful for cross-platform compatibility of builds with same flags.". In this particular case, the usecase is building libdeflate from a Rust wrapper, where I just want to use [`cc`](https://docs.rs/cc/1.0.58/cc/) with same flags (`-DFREESTANDING`) and have it built regardless of the underlying compiler.

Sure, I could detect is MSVC is targeted and pass some flags to it, but in this particular case it seems like a simpler change to just support this define on the `libdeflate` side in a cross-platform manner.

If you're strongly opposed to it (e.g. due to maintenance concerns), feel free to reject the PR.

>  Can you squash them together and write a correct commit message?

Yeah I just accidentally included that change and then pushed commit to revert. I thought you'll be just using Squash & Merge Github anyway, so it doesn't matter, but I can squash myself, sure.

> But actually compiler_msc.h still included <stdlib.h> in that version too (it's later in the file, not at the top).

Well, yeah, the order is what matters here, which is why this PR changes them.

ebiggers:
> I just want to use cc with same flags (-DFREESTANDING) and have it built regardless of the underlying compiler.

But you can't do this anyway since MSVC doesn't support the `-ffreestanding` option, right?  Or are you not actually doing a freestanding build with gcc or clang either, in which case what is your goal exactly?

The lib/utils.c change is a good change anyway, for robustness.  However it doesn't actually have anything to do with the commit "Depend on stdint.h instead of inttypes.h", but rather it's been that way since the initial "lib: add freestanding support".  Hence why I was confused.  Please squash together your changes and write a correct commit message.

For the change to make MSVC not use the built-in string functions: does MSVC have any equivalents that should be used instead?

lib/utils.c
```diff
@@ -25,15 +25,15 @@
...
 #ifdef FREESTANDING
 #  define malloc NULL
 #  define free NULL
 #else
 #  include <stdlib.h>
 #endif
 
-#include "lib_common.h"
-
 #include "libdeflate.h"
```
ebiggers, line 37: `libdeflate.h` should be moved above as well, in case it includes `stdlib.h` in the future.

RReverser:
> However it doesn't actually have anything to do with the commit "Depend on stdint.h instead of inttypes.h", but rather it's been that way since the initial "lib: add freestanding support".

What I meant is that original commit compiled successfully exactly because it didn't include `stdlib.h` after `malloc` and `free` redefinitions, but only `stddef.h`, so there was no conflict.

At least, when I spotted this syntax error and changed it back to `stddef.h` locally, then MSVC compiled it successfully, so I guess it's relevant?

If this is still confusing, please let me know how you would prefer to reword it :)

> For the change to make MSVC not use the built-in string functions: does MSVC have any equivalents that should be used instead?

MSVC defines `memcpy`, `memset` etc. as intrinsics themselves, which is why that conditional block basically ensures same behaviour as `__builtin_mem*` on GCC / Clang: https://docs.microsoft.com/en-us/cpp/intrinsics/intrinsics-available-on-all-architectures?view=vs-2019

> But you can't do this anyway since MSVC doesn't support the `-ffreestanding` option, right? Or are you not actually doing a freestanding build with gcc or clang either, in which case what is your goal exactly?

Yeah, that's a good point...

I probably will need to do compiler-specific flags after this anyway. For now I was mostly thinking about enforcing "set allocator before calling into libdeflate" in a consistent way across compilers, which is what `-DFREESTANDING` provides.

RReverser:
Btw, not in scope of this PR, but since you mentioned builtins... Is there any reason not to define `mem*` functions as

```diff
-#define memset(s, c, n)     __builtin_memset((s), (c), (n))
+#define memset              __builtin_memset
```

?

In theory, that would only make them more robust and closer to originals, since there is no token expansion involved for arguments.

ebiggers:
> What I meant is that original commit compiled successfully exactly because it didn't include stdlib.h after malloc and free redefinitions, but only stddef.h, so there was no conflict.

Nope, `stdlib.h` was still included after the definitions to NULL, so that version didn't work either.  Try it out for yourself...  The problem is in "lib: add freestanding support".

Anyway, I tried `/DFREESTANDING` with MSVC and it still doesn't work after this pull request because MSVC doesn't like the definitions of the string functions in `lib/utils.c` when they are intrinsic functions.  The `ifdef` condition there needs to be changed to exclude MSVC as well.

RReverser:
That's really weird. What version of MSVC are you using? I'm on 2019.

RReverser:
(sorry, got to go, will return to this PR tomorrow)

ebiggers:
> Btw, not in scope of this PR, but since you mentioned builtins... Is there any reason not to define `mem*` functions as
> 
> ```diff
> -#define memset(s, c, n)     __builtin_memset((s), (c), (n))
> +#define memset              __builtin_memset
> ```
> 
> ?
> 
> In theory, that would only make them more robust and closer to originals, since there is no token expansion involved for arguments.

As far as I know it doesn't actually matter, but if someone references `memset` without calling it, I think we want to give them a pointer to `memset` rather than `__builtin_memset`.  Hence the use of a function-like macro.  Also the gcc man page uses a function-like macro in its example.

None of the arguments are multiply-expanded, so there's no difference in behavior from the real version.

ebiggers:
> That's really weird. What version of MSVC are you using? I'm on 2019.

It's an older version, but it doesn't really matter.  If the build is not really a freestanding build, we shouldn't be either declaring or defining the string functions ourselves.

ebiggers:
Are you planning to do anything else with this pull request?  If not I'll just make the lib/utils.c change myself.

ebiggers:
> Are you planning to do anything else with this pull request? If not I'll just make the lib/utils.c change myself.

Made the lib/utils.c change in https://github.com/ebiggers/libdeflate/commit/303aecb074f1dd4ed8c897d94c3737cc58e84b92.

RReverser:
I'm sorry I didn't have time to return to this. Thanks for making the change!

RReverser:
Hmm, I think I'd still really like the `#if defined(FREESTANDING) && !defined(_MSC_VER)` to be merged as well.

Previously, you said

> But you can't do this anyway since MSVC doesn't support the `-ffreestanding` option, right? Or are you not actually doing a freestanding build with gcc or clang either, in which case what is your goal exactly?

which I agreed with, but now that I'm actually trying this, turns out that MSVC `cl.exe` just ignores unknown options including the `-ffreestanding`. Moreover, the Rust `cc` wrapper that I'm using has an API [`flag_if_supported`](https://docs.rs/cc/1.0.61/cc/struct.Build.html#method.flag_if_supported) that will only use a flag if it's supported by corresponding compiler.

If the change above would be merged, I could pass the same flags `-ffreestanding` + `-nostdlib` + `-DFREESTANDING` to `cc` crate, and it would build consistently as-is across all compilers, it's just that if user is on MSVC, it wouldn't be a truly freestanding build.

Does this make sense from a DX perspective?

RReverser:
(meanwhile, I can, of course, check whether the underlying compiler is MSVC and not pass flags when it's detected, but it seems that the change above is small enough and would help to simplify cross-platform builds)

ebiggers:
I'm still confused.  Why are you doing a freestanding build if you don't care whether it is actually freestanding?

RReverser:
I do want it to be truly freestanding when the target doesn't have a stdlib (such as wasm32 or some microcontroller targets), but the set of such targets isn't supported by MSVC anyway. 

As such, when the build script is executed on MSVC, I don't really care about it being truly freestanding or not, because the targets it's being used for already have stdlib.

All I want is to abstract away and cover both kinds of targets with a single script. For that, I either need the change above, or to replicate same condition and detect MSVC in my own build scripts. The change above seems like the path of the least resistance, and AFAIK doesn't have any obvious / major downsides.

PR #72, opened 2020-08-14
Make it (cross-)compile on Windows

thofma:


ebiggers:
`make install` (and `make uninstall` which you forgot to update) aren't intended to support Windows builds.  Why do you need them to work for Windows builds?

thofma:
Over at https://github.com/JuliaPackaging/Yggdrasil/pull/1449, we are cross-compiling libdeflate for various architectures and the build systems picks up the binaries and executables in `$prefix/bin` resp. `$prefix/lib`. So we rely on doing `make install` to install them at the right place.

ebiggers:
Are you sure that's the right thing to do?  Directories like `/usr/bin` aren't valid on Windows.

If you really have some use case, it's easy enough to support this.  But can you please fix `make uninstall` too, as I suggested?  Also, please make the commit message properly describe the change, like

    Makefile: support 'install' and 'uninstall' with Windows builds

Also, wouldn't you want to skip installing the `libdeflate.so` symlink too?  What does your build system do with it?

giordano:
Our use-case is to cross-compile _on_ GNU/Linux the binaries _for_ Windows, so `$(DESTDIR)$(BINDIR)` makes much sense for us.  Even if you don't care about this use-case, consistently using `PROG_SUFFIX` looks like a valuable change, pending doing the same for the `uninstall` rule.

PR #77, opened 2020-10-02
Improve architecture-specific matchfinder code.

astei:
This PR improves the architecture-specific matchfinder by adding dynamic dispatch, ala the existing Adler-32 and CRC32 paths. The primary goal is to allow more matchfinders to be added and improve performance somewhat.

The main impact is on x86: the AVX2 matchfinder can now be properly dynamically dispatched at runtime and if `-mavx2` is included in `CFLAGS` (or `-march` set to any platform with AVX2 support). On my Ryzen 9 3900X, I got an approximately 1% boost in deflate time (measured with a uncompressed tarball of the Silesia corpus) using just the changes in this PR and the regular CFLAGS, and a 2.7% boost when specifying `-mavx2` as `CFLAGS`. (I also tested with an Intel Xeon Skylake c5.large EC2 instance, and did not see any performance regression).

### DIFF
--- a/lib/x86/matchfinder_impl.h
+++ b/lib/x86/matchfinder_impl.h
@@ -25,13 +25,27 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
-#ifdef __AVX2__
+#include "cpu_features.h"
+
+#undef DISPATCH_MATCHFINDER_AVX2
+#if (!defined(DEFAULT_MATCHFINDER_INIT) || !defined(DEFAULT_MATCHFINDER_REBASE)) && \
+    (defined(__AVX2__) || (X86_CPU_FEATURES_ENABLED &&	\
+		COMPILER_SUPPORTS_AVX2_TARGET_INTRINSICS))
 #  if MATCHFINDER_ALIGNMENT < 32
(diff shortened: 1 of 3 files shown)


lib/x86/matchfinder_impl.h
```diff
@@ -25,13 +25,27 @@
...
 #  if MATCHFINDER_ALIGNMENT < 32
 #    undef MATCHFINDER_ALIGNMENT
 #    define MATCHFINDER_ALIGNMENT 32
 #  endif
+#  ifdef __AVX2__
+#    define ATTRIBUTES
+#    define DEFAULT_MATCHFINDER_INIT	matchfinder_init_avx2
+#    define DEFAULT_MATCHFINDER_REBASE	matchfinder_rebase_avx2
+#  else
+#    define ATTRIBUTES					__attribute__((target("avx2")))
```
astei, line 43: Note: this isn't compiling on the old Linux distribution test. I will fix this later.

ebiggers:
I ended up reworking this one too.  Can you take a look at https://github.com/ebiggers/libdeflate/pull/84?

I also merged https://github.com/ebiggers/libdeflate/pull/83, to improve testing of the dynamically-dispatched code.

PR #84, opened 2020-10-05
lib/matchfinder: support dynamic dispatch for init and rebase

ebiggers:


astei:
Hi! Just got some time to take a closer look at this and do some "proper" benchmarking (by ensuring there is little to no load on the CPU). Turns out this made a big difference...

Dynamic dispatch is actually a little bit worse than the current SSE2-only patch (but only by 1% by median), but this regression is probably acceptable and amortized throughout the run of a libdeflate-using program (the initialization only happens once). Forcing on the AVX2 path results in a 0.65% boost over the original SSE2-only path, and a ~1.67% boost over the dynamic dispatch.

System is still the same: AMD Ryzen 9 3900X, 32GB of DDR4-3600 CL18, Fedora 32 with GCC 10.2.1. My testing was done by running the included `benchmark` program over the Silesia corpus copied to a tmpfs beforehand over 30 runs and taking the median, so it probably leans towards disadvantaging dynamic dispatch.

PR #101, opened 2020-10-28
Add a CMakeLists.txt file

bsergean:
Hi there, thanks for this super useful and fast library.

I added a simple CMake file which should make it so that more users can use your library. One good recent feature of cmake is called FetchContent, which make it real easy to include another library by downloading it from github and building it. I added some doc about it in the README file.

### DIFF
--- a/libdeflate.h
+++ b/libdeflate.h
@@ -75,10 +75,35 @@ libdeflate_alloc_compressor(int compression_level);
 
 /*
  * libdeflate_deflate_compress() performs raw DEFLATE compression on a buffer of
- * data.  The function attempts to compress 'in_nbytes' bytes of data located at
- * 'in' and write the results to 'out', which has space for 'out_nbytes_avail'
- * bytes.  The return value is the compressed size in bytes, or 0 if the data
- * could not be compressed to 'out_nbytes_avail' bytes or fewer.
+ * data.  It attempts to compress 'in_nbytes' bytes of data located at 'in' and
+ * write the results to 'out', which has space for 'out_nbytes_avail' bytes.
+ * The return value is the compressed size in bytes, or 0 if the data could not
+ * be compressed to 'out_nbytes_avail' bytes or fewer (but see note below).
+ *
+ * If compression is successful, then the output data is guaranteed to be a
+ * valid DEFLATE stream that decompresses to the input data.  No other
+ * guarantees are made about the output data.  Notably, different versions of
+ * libdeflate can produce different compressed data for the same uncompressed
+ * data, even at the same compression level.  Do ***NOT*** do things like
+ * writing tests that compare compressed data to a golden output, as this can
+ * break when libdeflate is updated.  (This property isn't specific to
+ * libdeflate; the same is true for zlib and other compression libraries too.)
+ *
+ * Note: due to a performance optimization, libdeflate_deflate_compress()
+ * currently needs a small amount of slack space at the end of the output
+ * buffer.  As a result, it can't actually report compressed sizes very close to
+ * 'out_nbytes_avail'.  This doesn't matter in real-world use cases, and
+ * libdeflate_deflate_compress_bound() already includes the slack space.
+ * However, it does mean that testing code that redundantly compresses data
+ * using an exact-sized output buffer won't work as might be expected:
+ *
+ *	out_nbytes = libdeflate_deflate_compress(c, in, in_nbytes, out,
+ *						 libdeflate_deflate_compress_bound(in_nbytes));
+ *	// The following assertion will fail.
+ *	assert(libdeflate_deflate_compress(c, in, in_nbytes, out, out_nbytes) != 0);
+ *
+ * To avoid this, either don't write tests like the above, or make sure to
+ * include at least 9 bytes of slack space in 'out_nbytes_avail'.
  */
 LIBDEFLATEEXPORT size_t LIBDEFLATEAPI
 libdeflate_deflate_compress(struct libdeflate_compressor *compressor,


ebiggers:
Maintaining multiple build systems would be error-prone and extra work.  (And I only have time to work in libdeflate in my free time, so anything that avoids introducing additional maintenance burden would be very helpful.)  So the only way I might accept this is if CMake fully replaced the current Makefile.  That means the CMake replacement would need to support everything the current Makefile supports, and the README would need to be fully updated, and all scripts in the `scripts/` directory that are currently running `make` would need to be updated to use the CMake replacement instead.

Generally, my preference is to keep the plain Makefile.  FWIW, libdeflate actually used CMake originally, but pretty quickly I replaced it with a plain Makefile since it seemed just as simple, and it was easier to use (e.g. `make -j8 <options>` instead of `rm -r build; mkdir build; cd build; cmake .. <options>; make -j8` instead of just `make -j8`).  And I don't really buy claims like "CMake is becoming a standard now", as lots of projects instead use GNU autotools, a plain Makefile, Meson, Bazel, etc...

It's still possible that I could be convinced if there were large benefits to switching to CMake in particular, though, and if it actually replaced the current Makefile as mentioned above, and if you did a really good job with the changes.

bsergean:
Thanks for taking a look, you're making some good points.

I have a project where I still have a very tiny makefile in the top folder which takes care of invoking cmake for the common case which can be a chore. I do all my development on mac or linux.

I don't know at which pace I can improve this PR, but the current version is relatively complete, I'll look in more details at what is inside the scripts folder.

One advantage of cmake is that it is truly cross platform and works well on Windows, while autotools is not, and I haven't followed the latest trends, but bazel seems to be for niche projects or google only projects, same with buck (the facebook one). Not sure what big project uses Meson, I can't think of any, maybe jsoncpp. Also FetchContent is quite cool I think.

I know C++ more, but serious projects that now have cmake support are libcurl, libuv, spdlog, rapidjson, openssl. If your project has cmake it can be included more easily in vcpkg which is Microsoft attempt at adding a real package manager for C/C++. (there's something called cdeps too for just C).

bsergean:
I found something interesting where a build fail through cmake, but not with the regular make build. (the xenial warning config). The other outstanding thing I'm looking at is making shared libraries, where I get missing symbols.

```
In file included from /home/travis/build/bsergean/libdeflate/lib/adler32.c:63:0:

/home/travis/build/bsergean/libdeflate/lib/arm/adler32_impl.h: In function ‘adler32_neon_chunk’:

/home/travis/build/bsergean/libdeflate/lib/arm/adler32_impl.h:102:2: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]

  *s1 += v_s1[0] + v_s1[1] + v_s1[2] + v_s1[3];

  ^

/home/travis/build/bsergean/libdeflate/lib/arm/adler32_impl.h:103:2: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]

  *s2 += v_s2[0] + v_s2[1] + v_s2[2] + v_s2[3];

  ^

cc1: all warnings being treated as errors
```

ebiggers:
> ```
> In file included from /home/travis/build/bsergean/libdeflate/lib/adler32.c:63:0:
> 
> /home/travis/build/bsergean/libdeflate/lib/arm/adler32_impl.h: In function ‘adler32_neon_chunk’:
> 
> /home/travis/build/bsergean/libdeflate/lib/arm/adler32_impl.h:102:2: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
> 
>   *s1 += v_s1[0] + v_s1[1] + v_s1[2] + v_s1[3];
> 
>   ^
> 
> /home/travis/build/bsergean/libdeflate/lib/arm/adler32_impl.h:103:2: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
> 
>   *s2 += v_s2[0] + v_s2[1] + v_s2[2] + v_s2[3];
> 
>   ^
> 
> cc1: all warnings being treated as errors
> ```

That's a known false positive compiler warning with some gcc versions.  In `.travis.yml` (https://github.com/ebiggers/libdeflate/blob/master/.travis.yml#L43), this warning is suppressed on xenial + gcc + arm64, which is the only tested case where this warning is seen.

CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+target_compile_options(deflate PRIVATE ${lint-missing-prototypes})
+target_compile_options(deflate PRIVATE ${lint-strict-prototypes})
+target_compile_options(deflate PRIVATE ${lint-vla})
+target_compile_options(deflate PRIVATE ${lint-implicit-fallthrough})
+
+# The CFLAGS environment variable is read and handled automatically by cmake
+
+if (BUILD_SHARED_LIBS)
+  target_compile_options(deflate PRIVATE -fvisibility=hidden -D_ANSI_SOURCE)
+  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
SpaceIm, line 126: Not required. In CMake, shared libs always have `POSITION_INDEPENDENT_CODE ON`

CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+target_compile_options(deflate PRIVATE ${lint-declaration-after-statement})
+target_compile_options(deflate PRIVATE ${lint-missing-prototypes})
+target_compile_options(deflate PRIVATE ${lint-strict-prototypes})
+target_compile_options(deflate PRIVATE ${lint-vla})
+target_compile_options(deflate PRIVATE ${lint-implicit-fallthrough})
+
+# The CFLAGS environment variable is read and handled automatically by cmake
+
+if (BUILD_SHARED_LIBS)
+  target_compile_options(deflate PRIVATE -fvisibility=hidden -D_ANSI_SOURCE)
```
SpaceIm, line 125: There is an abstraction of visibility preset in CMake: https://cmake.org/cmake/help/latest/prop_tgt/LANG_VISIBILITY_PRESET.html

SpaceIm: Moreover `LIBDEFLATE_DLL` PUBLIC definition should be added if WIN32: https://github.com/ebiggers/libdeflate/blob/e1803dc145aaecd531fb081de2b9e354fe27fda7/Makefile.msc#L46

bsergean:
I updated the PR @Spacelm

bsergean:
```
$ mkdir build
$ cd build
$ cmake -DDEFLATE_BUILD_TEST_PROGRAMS=ON ..
$ make ...
$ make test
Running tests...
Test project /Users/benjaminsergeant/src/foss/github/libdeflate/build
    Start 1: test_checksums
1/6 Test #1: test_checksums ...................   Passed    0.25 sec
    Start 2: test_custom_malloc
2/6 Test #2: test_custom_malloc ...............   Passed    0.24 sec
    Start 3: test_incomplete_codes
3/6 Test #3: test_incomplete_codes ............   Passed    0.22 sec
    Start 4: test_litrunlen_overflow
4/6 Test #4: test_litrunlen_overflow ..........   Passed    0.24 sec
    Start 5: test_slow_decompression
5/6 Test #5: test_slow_decompression ..........   Passed    0.21 sec
    Start 6: test_trailing_bytes
6/6 Test #6: test_trailing_bytes ..............   Passed    0.21 sec

100% tests passed, 0 tests failed out of 6

Total Test time (real) =   1.37 sec
```

bsergean:
@ebiggers I've changed the PR to be purely additive, and not conflict with any existing makefiles or scripts.

I know you made your point about wanting to have a single build-system, but would you consider accepting the cmake files as community maintained, which can be highlighted even more in the readme? Project like libuv or curl have multiple build systems.

A bunch of folks use IDE projects generators, and having cmake support will let them build libdeflate when used in their project that way. Also the cmake ninja generator will work fine here.

rleigh-codelibre:
I was just looking at CMake support for libdeflate (for use with libTIFF), and noticed all the hard work had already been done!  This is really nice.

For building with Windows with MinGW or MSVC, CMake is a real improvement on the two Makefiles.  It would make it much easier to package and distribute with vcpkg for example.  The Makefile.msc has no install targets and no support for debug vs release runtimes.  CMake can provide all that directly.

Some comments on the PR:

* If libdeflate depends upon ZLIB, it would be nice to have a transitive dependency on `ZLIB::ZLIB` and have the exported configuration use `find_dependency` to automatically find and create the imported target.  You can see an example of this [here](https://gitlab.com/codelibre/ome/ome-files-cpp/-/blob/master/lib/ome/files/CMakeLists.txt#L226) L226-249 and [this template](https://gitlab.com/codelibre/ome/ome-files-cpp/-/blob/master/lib/ome/files/OMEFilesConfig.cmake.in).  Note: this was originally done with a fairly old CMake version; there may now be a more transparent way of importing the transitive dependencies.
* Shared library versioning doesn't look quite right; if the existing Makefiles don't do this either it would be nice to set up some proper versions rather than 0.0.0 for all build systems
* You might want to set the debug postfix to "d" for Windows, so we can have debug and release libraries coexisting (would also be nice for getting libdeflate packaged with vcpkg).  Example:
```
# Debug postfix
if(MSVC)
    set(CMAKE_DEBUG_POSTFIX "d")
endif()
```

As an aside, [FindDeflate](https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/FindDeflate.cmake) might be of interest.  The library naming doesn't quite match the static Makefile names; having the debug postfixes would be necessary to make this work properly.  If you can make the exported configuration stuff work properly, this can be removed, but just mentioning it in case it's of use to others who want to use libdeflate with CMake (as opposed to building with CMake).

Kind regards,
Roger

bsergean:
Thanks Roger, I’ll try to incorporate your feedback in the PR. 

1. I don’t think that the library depends on zlib, just the tests and not even sure
2. Getting the shared library versioning shouldn’t be too hard I’ll try to get that working
3. The debug postfix looks really easy to do.

Last thing is that we could put that in a repo maybe called libdeflate-cmake, similar to openssl ? I know that supporting multiple build system can be annoying. I think have a ‘community level’ support for CMake would be ideal.

Ps: we use (as opposed to build) libdeflate with CMake through ExternalProject version and it works alright.

SpaceIm:
I have created libdeflate recipe in conan public repository, and I'm also interested in CMake support in libdeflate. Indeed libdeflate is now an optional dependency of libtiff (since 4.2.0) and gdal (since 3.2.0), enabled by default, so it's indirectly included in the dependency graph of a lot of libraries. Therefore a robust build would be welcome.

ivan-volnov:
Any news here?

serge2016:
I am also very interested!

ckerr:
At the risk of piling on, I'd also like to see CMake support for libdeflate. It would make embedding it in my own CMake-based application much simpler. :)

ebiggers:
Given the interest, I'm willing to switch to CMake, but it has to replace the current Makefile instead of being added alongside it.

CMakeLists.txt
```diff
@@ -0,0 +1,164 @@
...
+        EXPORT deflate
+        ARCHIVE DESTINATION lib
+        PUBLIC_HEADER DESTINATION include/
+)
+
+# This is required to work with FetchContent
+install(EXPORT deflate
+        FILE deflate-config.cmake
+        NAMESPACE deflate::
+        DESTINATION lib/cmake/deflate)
```
SpaceIm, line 159: - Misleading comment, CMake config files are mainly for `find_package()`.
- Instead I would write a basic `deflate-config.cmake.in` file and rely on `CMakePackageConfigHelpers` module and its `configure_package_config_file`. I would export targets in `deflate-targets.cmake`. Do not forget to also export a target file in the build tree. A version file would be nice also (it requires to add VERSION to the project.
  Basically something like this:

  ```cmake
  configure_package_config_file(
      ${PROJECT_SOURCE_DIR}/cmake/deflate-config.cmake.in
      deflate-config.cmake
      INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/deflate
  )
  
  write_basic_package_version_file(
      deflate-config-version.cmake
      VERSION ${PROJECT_VERSION}
      COMPATIBILITY AnyNewerVersion
  )
  
  export(
      EXPORT deflate
      NAMESPACE deflate::
      FILE deflate-targets.cmake
  )
  
  install(
      EXPORT deflate
      NAMESPACE deflate::
      FILE deflate-targets.cmake
      DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/deflate
  )
  ```

- The makefile installs a pkgconfig file, it should be ported to the CMakeLists. The previous one can be reused.

SpaceIm (changes requested):

bsergean:
I think there should be a mini transition period where both makefile and cmakefiles exists, so that we can know that the new system actually works.

I'm still very interested in having this PR land, but I have also a lot of other things on my plate so I don't know when I'll get to this. @SpaceIm / how you could you add your proposed changes to this PR ? By making a PR in my repo ?

CMakeLists.txt
```diff
@@ -0,0 +1,164 @@
...
+
+install(TARGETS deflate
+        EXPORT deflate
+        ARCHIVE DESTINATION lib
+        PUBLIC_HEADER DESTINATION include/
+)
+
+# This is required to work with FetchContent
+install(EXPORT deflate
+        FILE deflate-config.cmake
```
eli-schwartz, line 157: The existing pkg-config export is called libdeflate.pc, used via
- `dependency('libdeflate')` (meson)
- `pkg_check_modules(LIBDEFLATE libdeflate IMPORTED_TARGET GLOBAL)` (cmake)
- `pkg-config --cflags --libs libdeflate` (Makefile)


Hence, the cmake-config name must be libdeflate-config.cmake, used via:
- `find_package(libdeflate)` (cmake)
- `dependency('libdeflate')` (meson)


It is extremely important the names be consistent between pkg-config and cmake. e.g. Meson can look up this project as a dependency using multiple methods: pkg-config, cmake, checking for [a standalone ini file](https://mesonbuild.com/Wrap-dependency-system-manual.html#wrap-format) defining the location to download the dependency from...

However, in order for that to work, the names *must* be the same, or Meson will look for `libdeflate-config.cmake` when it is strangely named `deflate-config.cmake`.

eli-schwartz (changes requested):

SpaceIm:
> @SpaceIm / how you could you add your proposed changes to this PR ? By making a PR in my repo ?

Sure.

eli-schwartz:
I commented on that PR. The generated pkg-config file has an edge case that leads to *extremely* broken behavior if the user configures with `-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib` (using an absolute path for values other than prefix).

This is very difficult to fix properly due to the design of cmake (although meson handles it perfectly), although there are ways to make it less terrible.

SpaceIm:
another one: https://github.com/bsergean/libdeflate/pull/2

MinGW is broken by the way. `-municode` must be injected during compilation of the the utility & tests, but I can't make it work with CMake.

jmarshall:
At the risk of piling on, but for the record: I for one would likely lose any interest in contributing to libdeflate if it switched from a Makefile to CMake.

ebiggers:
> At the risk of piling on, but for the record: I for one would likely lose any interest in contributing to libdeflate if it switched from a Makefile to CMake.

Can you elaborate on why that is?

jmarshall:
For me: decades of experience with Make, versus an opinion that CMake scripts are inflexible and obscure.

But really I just wanted to demonstrate that, notwithstanding all the CMake advocates posting on this PR, there do exist people who would prefer the status quo.

But no great loss: I haven't yet contributed to libdeflate in any case. 😃

ivan-volnov:
> At the risk of piling on, but for the record: I for one would likely lose any interest in contributing to libdeflate if it switched from a Makefile to CMake.

> But no great loss: I haven't yet contributed to libdeflate in any case. 😃

Nice try ))

Well, I personally believe that CMake is very important for many developers. CMake support allows easy integration with current big projects that use the library.

Even Boost, having their own build system, is gradually switching to CMake.

So, in my opinion, the library can gain a lot of contemporary developers by only switching the build system.

eli-schwartz:
(Boost's own build system also has the curious status of being *worse* than handwritten Makefiles.)

SunBlack:
> For me: decades of experience with Make, versus an opinion that CMake scripts are inflexible and obscure.

Seems you have never really worked on bigger projects which are multiplatform otherwise you would prefer CMake over Makefile:
- Generation for various IDEs (there are still some projects which provides solutions for each MSVC version in their repo)
- You don't need to search the whole Makefiles for available options (not all are adding a documentation at top of the Makefile)
- Generation of installer or deb files (last thing would be e.g. nice to have here)
- Checking of available features (in this MR I missing e.g. the AVX detection, this could resolve the issue, that the MSVC build result is slower than this of mingw)
- Easy integration within other projects via imported targets. I don't know e.g. how do check the minimum required version with a Makefile.
- ...

Of course CMake is not perfect (as it is a scripting language), therefore tools like CMake format are not even perfect. But compared to using plain Makefiles it is a dream (if you are using CMake 3.x) ;-)

eli-schwartz:
- not everyone cares about or likes IDEs, but you seem to be talking more about MSVC than about IDEs?

- Makefiles aren't a real build system, real build systems like autotools and meson have `--help` or `configure` commands that do this for you. Sadly, this is actually a point against cmake, since cmake is like raw Makefiles and requires reading the entire build files to find undocumented variables that can be and usually are anywhere. Cmake doesn't have a help command. You can run a command ***after*** you unsuccessfully configured the project with the wrong options, that prints every internal state variable whether they are options or not, without context or guidance, though.

- Generation of deb files is an antipattern, please use the distro tooling or something like `checkinstall`. :)

- again Makefiles aren't a real build system. Real build systems like autotools and meson and cmake too, lets you check for features before building. It's hardly unique to cmake, but you can certainly do it with cmake.

- Integration with other projects via imported targets is an antipattern, pkg-config exists and has the benefit of working with projects that aren't cmake. That being said, you can install cmake config files from any build system, it doesn't require you to build with cmake.
  Also yes, if you already use cmake you can add another cmake project using one of 3 different verbose and clunky methods which all break picking up system copies without even more scripting. At least you can do it! You cannot do it at all with raw Makefiles.

...

Basically you're making a lot of arguments in favor of having a build system at all. I agree. Real build systems are good. ~~Even cmake is better than no build system at all.~~

jmarshall:
I wrote a brief comment as a counterpoint to similarly content-free brief comments from CMake advocates, in response to https://github.com/ebiggers/libdeflate/pull/101#issuecomment-718326339. I did not wish to derail this PR or add to @ebiggers's workload, so I have deleted my comments.

I suggest @SunBlack refrain from making baseless & incorrect comments on other people's experience and expertise.

SunBlack:
> * not everyone cares about or likes IDEs, but you seem to be talking more about MSVC than about IDEs?

Not only, but as this repo contains a `Makefile.msc`, so you usually use MSVC when using this. But of course you could also use Qt Creator, Eclipse or some other IDEs.

> * Makefiles aren't a real build system, real build systems like autotools and meson have `--help` or `configure` commands that do this for you. Sadly, this is actually a point against cmake, since cmake is like raw Makefiles and requires reading the entire build files to find undocumented variables that can be and usually are anywhere. Cmake doesn't have a help command. You can run a command _**after**_ you unsuccessfully configured the project with the wrong options, that prints every internal state variable whether they are options or not, without context or guidance, though.

True, going through the configure step of CMake is more like going through a Wizard than reading the man page before using the CLI. Nevertheless you forgot: configure is again just a typical script on Unix platforms. So you need again another script for non Unix platforms.

> * Generation of deb files is an antipattern, please use the distro tooling or something like `checkinstall`. :)

Why this is an antipattern? When using `checkinstall`  you need to add everything yourself, which is not standard (e.g. required data files). Therefore you have always remember which extra information you have to add (usually you store the last used call somewhere). The integration of CMake doesn't need it - this can be even dynamic depends on your build selection. Another good thing: You have to define most information just once and not multiple times. E.g. information like the version, company,... are stored in a header file (to show it in the program itself), DLLs (As far as I know Linux doesn't show this information on a .so file) and installer. So you have to adjust just one single point to adjust it everywhere.

> * again Makefiles aren't a real build system. Real build systems like autotools and meson and cmake too, lets you check for features before building. It's hardly unique to cmake, but you can certainly do it with cmake.

Yep, that why play Makefiles are sth. you used 10 years ago. As software is getting more complex and you usually sometimes want to enable or disable specific hardware features, Makefiles aren't the best anymore for it (even it is still possible)

> * Integration with other projects via imported targets is an antipattern, pkg-config exists and has the benefit of working with projects that aren't cmake. That being said, you can install cmake config files from any build system, it doesn't require you to build with cmake.

pkg-config is usually just Unix (never saw it someone using on Windows as there is vcpkg more common) and to have lines like `pkg-config --libs --cflags glib-2.0` within your Makefile is not really the best in terms of readability (when you starting to make the version configurable).

> Basically you're making a lot of arguments in favor of having a build system at all. I agree. Real build systems are good. ~Even cmake is better than no build system at all.~

Yep that is the point, as the current integration of libdeflate makes no fun.

SpaceIm:
> Integration with other projects via imported targets is an antipattern, pkg-config exists and has the benefit of working with projects that aren't cmake.

It's not an anti-pattern. C++ ecosystem is fragmented, there is no official tool. Some libraires provide CMake config files & pkgconfig files (and most of the time it's broken in complex libraries), some other just one flavor, some other a custom shell config file, some other don't even care. I always advice CMake imported targets over pkgconfig files in a CMake based project, it's usually more robust.

bsergean:
I just closed this, this is where the work is happening now -> https://github.com/ebiggers/libdeflate/pull/235

ebiggers:
The master branch is using CMake now.  I'd appreciate it if people would verify that it's working properly for their use cases.

bsergean:
I just tried to build master with ExternalProject, and it works fine.

```
ExternalProject_Add(
  project_deflate
  EXCLUDE_FROM_ALL TRUE
  PREFIX ${EXTERNAL_PROJECTS_DIR}
  GIT_REPOSITORY "https://github.com/ebiggers/libdeflate.git"
  GIT_TAG master
  CMAKE_ARGS -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
)
```

diizzyy:
Basic compile and unit tests works as expected on FreeBSD 13.1-STABLE

ckerr:
> The master branch is using CMake now. I'd appreciate it if people would verify that it's working properly for their use cases.

@ebiggers 03fba38 is working for my project :+1:. LGTM

rleigh-codelibre:
I created https://github.com/ebiggers/libdeflate/issues/236 after some initial testing on Windows.  There is a portability issue with the shared/static library naming which will need fixing to make the software buildable and usable on Windows.

PR #112, opened 2020-12-22
programs/prog_util.h: include sys/types.h

ffontaine:
Include `sys/types.h` to avoid the following build failure on uclibc:

```
In file included from programs/gzip.c:28:0:
programs/prog_util.h:159:1: error: unknown type name ‘ssize_t’
 ssize_t xread(struct file_stream *strm, void *buf, size_t count);
 ^
```

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/programs/prog_util.h
+++ b/programs/prog_util.h
@@ -38,6 +38,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifndef _WIN32
+#  include <sys/types.h>
+#endif
 
 #include "../common/common_defs.h"
 


ebiggers:
Can it be `<sys/types.h>` instead?  It's better to use a more specific header.  Also that's needed is the definition of `ssize_t`.

Also, please rebase onto the latest `master` branch.  That should fix the test failures.

ebiggers:
This pull request now includes a merge commit; can you rebase it so that it's just the one commit that actually makes the change, on top of the latest `master` branch?  Thanks!

ffontaine:
Done

ebiggers:
Merged, thanks.

PR #115, opened 2021-01-13
Define Wasm as fast unaligned access

RReverser:
I saw this on Twitter - https://twitter.com/Algunenano/status/1317098341377900550 - claiming this flag to make libdeflate run 20% faster on Wasm.

Indeed, when tried even in a PNG compression benchmark I've observed 10-15% improvement when this flag is enabled.

Even though Wasm might be running on top of a variety of underlying platforms, the spec requires it to support unaligned access, and on majority of platforms it will translate to a faster code. Hence, I think it makes sense to enable this flag by default.

### DIFF
--- a/common/compiler_gcc.h
+++ b/common/compiler_gcc.h
@@ -175,7 +175,16 @@ typedef unsigned char      __v32qu __attribute__((__vector_size__(32)));
 #  define bswap64	__builtin_bswap64
 #endif
 
-#if defined(__x86_64__) || defined(__i386__) || defined(__ARM_FEATURE_UNALIGNED) || defined(__powerpc64__)
+#if defined(__x86_64__) || defined(__i386__) || \
+    defined(__ARM_FEATURE_UNALIGNED) || defined(__powerpc64__) || \
+    /*
+     * For all compilation purposes, WebAssembly behaves like any other CPU
+     * instruction set. Even though WebAssembly engine might be running on top
+     * of different actual CPU architectures, the WebAssembly spec itself
+     * permits unaligned access and it will be fast on most of those platforms,
+     * and simulated at the engine level on others, so it's worth treating it
+     * as a CPU architecture with fast unaligned access.
+    */ defined(__wasm__)
 #  define UNALIGNED_ACCESS_IS_FAST 1
 #endif
 


ebiggers:
Can you explain why the existing checks for architectures that support fast unaligned access aren't sufficient?

I.e. this change needed because unaligned access support for x86, arm, or powerpc wasn't being detected by the existing code, or is it working fine on those architectures but you want it to be detected for other architectures too?

Please include a brief explanation in a comment and not just in the commit message, so that people reading the code don't need to dig through commit messages.

Also, please line wrap the code at 80 columns and the commit message at 72 columns.

RReverser:
> is it working fine on those architectures but you want it to be detected for other architectures too

It's working fine on those architectures, but not on WebAssembly - it was using the slower path - so I added it to the existing list.

> Please include a brief explanation in a comment and not just in the commit message, so that people reading the code don't need to dig through commit messages.

Hmm, what's special / different about WebAssembly that requires an explanation while x64 / ARM / etc. on the same line don't have such comment?

ebiggers:
> > is it working fine on those architectures but you want it to be detected for other architectures too
> 
> It's working fine on those architectures, but not on WebAssembly - it was using the slower path by default - so I added it to the existing list.
> 
> > Please include a brief explanation in a comment and not just in the commit message, so that people reading the code don't need to dig through commit messages.
> 
> Hmm, what's special / different about WebAssembly that requires an explanation while x64 / ARM / etc. on the same line don't have such comment? Wasm check is no different from all the other target platform checks on the same line.

It's pretty well known that x86, ARM, etc. are CPU architectures.  It's less well know that WebAssembly also behaves like a CPU architecture (at least, I didn't know this... but I don't think I'm alone).

RReverser:
> It's less well know that WebAssembly also behaves like a CPU architecture (at least, I didn't know this... but I don't think I'm alone).

Interesting! Thanks for sharing your perspective. I never realised it can be seen differently than other platforms / instruction sets, but, admittedly, due to the nature of my job I'm pretty deep in the WebAssembly world so I might be the odd one out.

I'll update the PR tomorrow.

RReverser:
Done.

ebiggers:
Merged, thanks.

Issue #125, opened 2021-03-15
Test file integrity option


tansy:
I noticed that gzip doesn't have option for testing file integrity so I have done one. It's not pull request, but diff ready for patching. It may not be very you would do that but have a look and consider to add the option as it's extremely useful and simple to implement in the same time.

Reason of this request is that libdeflate is definitely fastest implementation of all known to me and when it comes to checking of big amount of gzip compressed data it is my favourite but sending it to /dev/null is diminishing benefits.

I attach diff file to the post.

[libdeflate-1_7-testa_diff.GZ](https://github.com/ebiggers/libdeflate/files/6144975/libdeflate-1_7-testa_diff.GZ)

ebiggers:
At first glance, this seems reasonable to support.  GNU gzip and gunzip has this option too, and it doesn't appear to require any changes to the library.  (It would be faster with library support, but it doesn't require it; just avoiding writing the data to a file descriptor is still a benefit.)

Can you open this as a pull request so that it can be properly reviewed?

tansy:
Sure, I'll try. I'm not fan of git as I don't know it. That's why I attached diff patch. Give me few days, I have to find time.

PR #127, opened 2021-03-24
adding test file integrity option

tansy:
As mentioned in issue #125 this pr adds file integrity test option to gzip client.


tansy:
I'm not sure what happened. Shall I reopen this request?

ebiggers:
If you want this to be reviewed and merged, then yes you need to reopen it.

programs/gzip.c
```diff
@@ -43,11 +43,12 @@ struct options {
 	bool decompress;
 	bool force;
 	bool keep;
+	bool test;
 	int compression_level;
 	const tchar *suffix;
 };
 
-static const tchar *const optstring = T("1::2::3::4::5::6::7::8::9::cdfhknS:V");
+static const tchar *const optstring = T("1::2::3::4::5::6::7::8::9::cdfhkntS:V");
```
ebiggers, line 51: Please keep the options in alphabetical order.

programs/gzip.c
```diff
@@ -65,6 +66,7 @@ show_usage(FILE *fp)
 "  -f        overwrite existing output files\n"
 "  -h        print this help\n"
 "  -k        don't delete input files\n"
+"  -t        test file integrity\n"
```
ebiggers, line 69: Please keep the options in alphabetical order.

programs/gzip.c
```diff
@@ -258,7 +260,13 @@ do_decompress(struct libdeflate_decompressor *decompressor,
 			goto out;
 		}
 
-		ret = full_write(out, uncompressed_data, actual_out_nbytes);
+		if (out != NULL) {
+			ret = full_write(out, uncompressed_data, actual_out_nbytes);
+		}
+		else { /// NULL means test mode: don't write output
+			ret = uncompressed_size; /// as do_decompress() finished successfully it means file was decompressed fine
```
ebiggers, line 267: This doesn't make sense; this is making the program always fail (exit with failure status) when `-t` was specified.

I recommend passing the `options` to this function.  Then it would be pretty self-explanatory to write:

                                                                                 
                if (!options->test) {                                            
                        ret = full_write(out, uncompressed_data,                 
                                         actual_out_nbytes);                     
                        if (ret != 0)                                            
                                goto out;                                        
                }                                                                


In addition to fixing this, can you please add test cases to `scripts/gzip_tests.sh` to test that this new option works correctly?

programs/gzip.c
```diff
@@ -416,6 +424,9 @@ decompress_file(struct libdeflate_decompressor *decompressor, const tchar *path,
 	if (ret != 0)
 		goto out_close_in;
 
+	if (!options->test) { /// indicate empty output file name, mock use of stdout
```
ebiggers, line 427: `options->to_stdout` is always set when `options->test`, which implies `newpath == NULL` already.  So this part is unnecessary.

programs/gzip.c
```diff
@@ -425,7 +436,10 @@ decompress_file(struct libdeflate_decompressor *decompressor, const tchar *path,
 	if (ret != 0)
 		goto out_close_out;
 
-	ret = do_decompress(decompressor, &in, &out);
+	if (options->test) /// decompress file, if it succeeds, test went ok, if fails, library generates message anyways
```
ebiggers, line 439: It would be more self-explanatory to just pass the `options` down to `do_decompress()`.

`ret = do_decompress(decompressor, &in, &out, options);`

programs/gzip.c
```diff
@@ -534,6 +548,7 @@ tmain(int argc, tchar *argv[])
 	options.decompress = is_gunzip();
 	options.force = false;
 	options.keep = false;
+	options.test = false; /// don't test by default
```
ebiggers, line 551: This comment isn't helpful, as this line is self-explanatory.

programs/gzip.c
```diff
@@ -576,6 +591,11 @@ tmain(int argc, tchar *argv[])
 			 *  option as a no-op.
 			 */
 			break;
+		case 't': /// test
+			options.test = true;
+			options.decompress = true;
+			options.to_stdout = true;
```
ebiggers, line 597: It's not obvious why `to_stdout` is being set here.  After all, `-t` doesn't actually write anything to standard output.  It would be helpful to add a comment that properly explains this:

                        /*                                                       
                         * -t behaves just like the more commonly used -c        
                         * option, except that -t doesn't actually write         
                         * anything.  For ease of implementation, just pretend   
                         * that -c was specified too.                            
                         */                                                      
                        options.to_stdout = true;                                

Either that or properly abstract this out into something like `options.no_output_file`.  I'm leaning towards just reusing `to_stdout` as you've done, but it needs to be properly explained.

programs/gzip.c
```diff
@@ -605,7 +625,7 @@ tmain(int argc, tchar *argv[])
 	}
 
 	ret = 0;
-	if (options.decompress) {
+	if (options.decompress || options.test) {
```
ebiggers, line 628: This part is unnecessary, as `test` implies `decompress`.

programs/gzip.c
```diff
@@ -576,6 +591,11 @@ tmain(int argc, tchar *argv[])
 			 *  option as a no-op.
 			 */
 			break;
+		case 't': /// test
```
ebiggers, line 594: Please keep the options in alphabetical order.

ebiggers (changes requested):

programs/gzip.c
tansy: Sure

programs/gzip.c
tansy: Sure

programs/gzip.c
tansy: I didn't know what exactly is to be returned.
That's also why I included diff patch rather than pull request.

programs/gzip.c
tansy: Understood.

tansy:
Comments made were redundant for a reason, I wanted to explain what it actually do, so they were for you, and it's true they were excessive.

And yes, I wanted to utilise `to_stdout` option to not contrive new mechanisms.

programs/gzip.c
tansy: That would mean adding one argument to this function.

tansy:
I have changed that. Looks better now. Can you check this [paste](https://pastebin.com/D9DKyawx) for the time being?

I didn't do test script yet, and there is a question - do we need verbose mode (`-v`) for test like gzip?

ebiggers:
> I have changed that. Looks better now. Can you check this [paste](https://pastebin.com/D9DKyawx) for the time being?
> 
> I didn't do test script yet, and there is a question - do we need verbose mode (`-v`) for test like gzip?

No need for pastebin; just update this pull request.

I don't see how `-v` is related.  Wouldn't that be a separate addition?  It could be added later if someone wants it.

tansy:
> I don't see how -v is related

As it's associated feature.

```
$  gzip -t fox2-1.gz fox2-bad-flip.gz 
gzip: fox2-bad-flip.gz: invalid compressed data--crc error
```
vs
```
$ gzip -tv fox2-1.gz fox2-bad-flip.gz 
fox2-1.gz:       OK
fox2-bad-flip.gz:
gzip: fox2-bad-flip.gz: invalid compressed data--crc error
```
Without it gzip doesn't explicitly state the tested file is ok. Only returns 0 on success, which is always.
libdeflate-gzip, as it is now, returns 0 when file is correct and 1 if not. This bevaviour is inherited form way it decompresses - it returns 1 if file if incorrect and fails to decompress.

tansy:
There is corrected, [new version](https://github.com/ebiggers/libdeflate/pull/127/commits/a754af276c280a6d0eb88d9fb13755c4e04a2ef6) with testing script. Script has to be added to `gzip_tests.sh` in place and way you decide. I didn't do that as it's, I mean gzip_tests.sh, magical enough to not know where exactly should it be.

programs/gzip.c
```diff
@@ -258,9 +261,11 @@ do_decompress(struct libdeflate_decompressor *decompressor,
...
 		}
 
-		ret = full_write(out, uncompressed_data, actual_out_nbytes);
-		if (ret != 0)
-			goto out;
+		if (!options->test) {
+			ret = full_write(out, uncompressed_data, actual_out_nbytes);
+			if (ret != 0)
+				goto out;
+		} /* "else" (test): as do_decompress() finished successfully it means file was decompressed properly */
```
ebiggers, line 268: This comment is wrong since `do_decompress()` hasn't finished successfully yet.

I don't think it adds anything; it would be better to just remove this comment.

programs/gzip.c
```diff
@@ -583,6 +589,15 @@ tmain(int argc, tchar *argv[])
 				return 1;
 			}
 			break;
+		case 't':
+			options.test = true;
+			options.decompress = true;
+			options.to_stdout = true;
+			/*
+			 *  test utilises to_stdout (-c) option
+			 *  except that it doesn't write anything out
```
ebiggers, line 598: This doesn't really explain why `to_stdout` is set.  I recommend using the comment that I suggested.

programs/gzip.c
```diff
@@ -605,7 +620,7 @@ tmain(int argc, tchar *argv[])
 	}
 
 	ret = 0;
-	if (options.decompress) {
+	if (options.decompress) { /* also test */
```
ebiggers, line 623: I don't think this comment adds anything, as test implies decompress already.

scripts/test_integrity-b.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/sh
+#
+# test integrity of few basic files
```
ebiggers, line 3: This needs to be wired up with `gzip_tests.sh`, otherwise it won't run in CI.

I think you're overthinking this a bit.  All we need are some quick tests in `gzip_tests.sh`.  Something like the following:

```
begin_test '-t (test) option works'                                              
good_files=(                                                                     
'H4sIAAAAAAAAA3PMSVTITVTIzi9JVABTIJ5jzpGZelwAX+86ehsAAAA='                       
'H4sIAAAAAAAAAwvJSFUoLM1MzlZIKsovz1NIy69QyCrNLShWyC9LLVIoAUrnJFZVKqTkp+txAQBqzFDrLQAAAA==')
bad_files=(                                                                      
'H4sIAO1YYmAAA3PMSVTITVTIzi9JVABTIJ5jzpGZelwAX+46ehsAAAA='                       
'H4sIAO1YYmAAA3PMSVTITVTIzi85VABTIJ5jzpGZelwAX+86ehsAAAA='                       
'H4sIAAAAAAAAA3PMSVTITVTIzi9JVABTIJ5jzpGZelwAX+86ehsBAAA='                       
'H4sIAAAAAAAAAwvJSFUoLM1MzlZIKsovz1NIy69QyCrNLShWyC9LLVIogUrnJFZVKqTkp+txAQBqzFDrLQAAAA=='
'H4sIAAAAAAAAAwvJSFUoLM1MzlZIKsovz1NIy69QyCrNLShWyC9L')                          
for contents in "${good_files[@]}"; do                                           
        echo "$contents" | base64 -d | gzip -t                                     
done                                                                             
for contents in "${bad_files[@]}"; do                                            
        echo "$contents" | base64 -d > file                                        
        assert_error '\<invalid compressed data|file corrupt|unexpected end of file\>' \
                gzip -t file                                                     
done  
```                                                                           

Note that `gzip_tests.sh` runs with bash, so it doesn't need to be POSIX compliant (it looks like you ran into problems since you used `#!/bin/sh` instead of `#!/bin/bash`).  It just needs to pass shellcheck.

ebiggers (commented):
Sorry for the slow review!  I left a few more comments; the main remaining thing is to properly wire up the tests.

ebiggers:
> Sorry for the slow review! I left a few more comments; the main remaining thing is to properly wire up the tests.

Also, please squash your commits into one commit.  There's no need to keep every incremental change to this pull request as its own commit.

programs/gzip.c
tansy: You're right, I meant `libdeflate_gzip_decompress_ex()`, which returns `result`.

Issue #21, opened 2016-11-04
Compile failures with clang-3.7 and icc 16


nemequ:
https://travis-ci.org/quixdb/squash/builds/173186455

I checked clang-3.7 on an Ubuntu 14.04 VM and the errors happen with the Makefile included in libdeflate, too.

nemequ:
If you want, I'd be willing to put together a Travis CI config file for libdeflate.  That would make it easier to spot these issues and test fixes…

ebiggers:
Thanks, the problem with clang 3.7 was that clang 3.7 introduced the 'target' function attribute, but target-specific intrinsics can't be used in 'target' attribute functions until clang 3.8.  I worked around this by checking the clang version number (unfortunately I could not think of a better way).

icc may have a similar problem but I don't have it available to test right now.

I'm definitely interested in improving how libdeflate is tested, so feel free to submit a proposal.  Do note that I don't want to tie the project too closely with 3rd party services like Github and Travis CI, but in general I am all for better testing.  Note that I recently added my own test script (see `tools/run_tests.sh`).

nemequ:
For the undefined identifiers, including `immintrin.h` did the trick.  I just added `-include immintrin.h` to my cflags, but I assume there is a reasonable place in the code to do it.

Once that was fixed I started seeing

```
lib/decompress_impl.h(36): warning #3550: target attribute not recognized
  static enum libdeflate_result ATTRIBUTES
                                ^
```

ICC just doesn't support the bmi2 attribute.  It also doesn't have `__builtin_ia32_pdep_di`.  I fixed it with

```diff
diff --git a/common/compiler_gcc.h b/common/compiler_gcc.h
index 6b17dda..f1fec3e 100644
--- a/common/compiler_gcc.h
+++ b/common/compiler_gcc.h
@@ -63,7 +63,7 @@
 
 #define COMPILER_SUPPORTS_BMI2_TARGET                          \
        (COMPILER_SUPPORTS_TARGET_FUNCTION_ATTRIBUTE &&         \
-        (GCC_PREREQ(4, 7) || __has_builtin(__builtin_ia32_pdep_di)))
+        ((!defined(__INTEL_COMPILER) && GCC_PREREQ(4, 7)) || __has_builtin(__builtin_ia32_pdep_di)))
 
 #define COMPILER_SUPPORTS_AVX2_TARGET                          \
[...]

PR #130, opened 2021-05-05
fix ICC compilation

cielavenir:
now icc oneapi is freeware, so we can test it.

related to #21 

- on ICC, __v2di is defined in immintrin.h

```
In file included from lib/x86/crc32_impl.h(52),
                 from lib/crc32.c(186):
lib/x86/crc32_pclmul_template.h(107): error: identifier "__v2di" is undefined
  	const __v2di multipliers_4 = (__v2di){ 0x8F352D95, 0x1D9513D7 };
```

- __v64qi etc are not available on ICC

```
In file included from lib/adler32.c(65):
lib/x86/adler32_impl.h(114): error: identifier "__v64qi" is undefined
  	const __v64qi multipliers = (__v64qi){
```

tested with `icc (ICC) 2021.2.0 20210228`

### DIFF
--- a/common/compiler_gcc.h
+++ b/common/compiler_gcc.h
@@ -147,7 +147,8 @@
  * needed.
  */
 #if (GCC_PREREQ(4, 0) && !GCC_PREREQ(5, 1)) || \
-    (defined(__clang__) && !CLANG_PREREQ(3, 9, 8020000))
+    (defined(__clang__) && !CLANG_PREREQ(3, 9, 8020000)) || \
+    defined(__INTEL_COMPILER)
 typedef unsigned long long  __v2du __attribute__((__vector_size__(16)));
 typedef unsigned int        __v4su __attribute__((__vector_size__(16)));
 typedef unsigned short      __v8hu __attribute__((__vector_size__(16)));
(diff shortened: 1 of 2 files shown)


lib/x86/adler32_impl.h
```diff
@@ -29,6 +29,17 @@
 #define LIB_X86_ADLER32_IMPL_H
 
 #include "cpu_features.h"
+#ifdef __INTEL_COMPILER
+typedef unsigned int __v4su __attribute__ ((__vector_size__ (16)));
```
ebiggers, line 33: There is already a code block in `common/compiler_gcc.h` that defines missing vector types for gcc and clang when needed.  Could you put this in the same place, and reuse code if possible?

ebiggers (commented):
Can you please squash your commits together and include the explanation in the commit message, not the pull request description?  The pull request description doesn't get included in the git history.

One other comment below.

cielavenir:
squashed.

detailness of the commit message is arguable as full error message is verbose.

also, should I enable GCC_PREREQ on __INTEL_COMPILER? It looks like enabling is safe if we use ICC 2021.2. I cannot test older ICC though (only oneapi versions are freeware).

lib/x86/adler32_impl.h
cielavenir: done

ebiggers:
> also, on compiler_gcc.h, should I enable GCC_PREREQ on __INTEL_COMPILER? It looks like enabling is safe if we use ICC 2021.2. I cannot test older ICC though (only oneapi versions are freeware).

Do you mean change GCC_PREREQ to also return true if the compiler is icc?  I don't think that should be done, since GCC_PREREQ is meant specifically for gcc, not for clang or icc.  icc should be handled specially (when applicable) in places where the compiler version is being checked.

ebiggers:
Merged, thanks!

PR #135, opened 2021-08-28
Add DONTBUILD_SHARED_LIBS option

SCOTT-HAMILTON:
Hello !

I'm trying to build Qt statically on Nixpkgs. This implies that all it's direct and indirect dependencies are also built statically.
The way nixpkgs' static building works implies that no shared libs are built (because it forces -static option).

And qtbase depends on gdk-pixbuf which depends on libtiff which depends on libdeflate cf https://gitlab.com/libtiff/libtiff/-/blob/master/ChangeLog#L1136

(There is also qtbase which depends on cups which depends on libtiff)

This PR is there to avoid technical debt implied in maintaining general purpose options like this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,6 +212,7 @@ $(STATIC_LIB):$(STATIC_LIB_OBJ)
 DEFAULT_TARGETS += $(STATIC_LIB)
 
 # Create shared library
+ifndef DONTBUILD_SHARED_LIBS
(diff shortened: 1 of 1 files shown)


ebiggers:
I merged https://github.com/ebiggers/libdeflate/pull/137 instead, which accomplishes the same thing.

PR #137, opened 2021-11-04
Add environment variable to disable building shared library

KAction:
Environment variable DISABLE_SHARED (following convention of --disable-shared
of ./configure script) disables building of shared library and shared lib
symlink. It makes life of downstream maintainer easier when maintaining package
for environment that supports only static libraries.

See https://github.com/NixOS/nixpkgs/pull/144438


ebiggers:
Looks good!  Merged, thanks.

Issue #147, opened 2021-12-30
support cross-compile


leleliu008:
I wanna cross-compile this lib for Android.

I suggest define a var, say `TARGET_OS`, user can override it, default value is `uname -s`

https://github.com/ebiggers/libdeflate/blob/master/Makefile#L116

ebiggers:
You can already build it for Android.  Just point `CC` to the appropriate compiler from the Android NDK, e.g.:

```
make CC=$HOME/android-ndk-r23b/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android29-clang
```

leleliu008:
don't work on macOS. Build machine must be GNU/Linux.

ebiggers:
It should work; what is the error message you get?  (Obviously, you'd need to replace `linux` in the NDK path.)

Note, I don't have a macOS machine to test on.

ebiggers:
Probably the problem is the line that detects macOS builds:

```
else ifeq ($(shell uname),Darwin)
```

It should check the output of `$(CC) -dumpmachine` instead, like is done to detect Windows builds with MinGW.

A pull request would be appreciated, as I don't have a macOS machine to test on.

leleliu008:
```patch
diff --git a/Makefile b/Makefile
index 780d99f..948ab92 100644
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ ifneq ($(findstring -mingw,$(shell $(CC) -dumpmachine 2>/dev/null)),)
     endif

 # macOS?
-else ifeq ($(shell uname),Darwin)
+else ifeq ($(findstring -darwin,$(shell $(CC) -dumpmachine 2>/dev/null)),-darwin)
    SHARED_LIB         := libdeflate.$(SOVERSION).dylib
    SHARED_LIB_SYMLINK := libdeflate.dylib
    SHARED_LIB_CFLAGS  := -fPIC
```

this patch works for me.

do you accept github actions to test Android platform? here is a axample https://github.com/leleliu008/test/actions/runs/1644009507

I suggest using the Modern build system like CMake, meson, etc to make this project more popular.

ebiggers:
> do you accept github actions to test Android platform? here is a axample https://github.com/leleliu008/test/actions/runs/1644009507

[...]

PR #150, opened 2022-01-01
buildsys: change verify macOS platform condition

leleliu008:


ebiggers:
Can you use `ifneq`, so it's like the MinGW check?

    else ifneq ($(findstring -darwin,$(shell $(CC) -dumpmachine 2>/dev/null)),)

Also, can you please update the commit message to explain what this is fixing, and add `Fixes #147` to link it to the issue you filed?

Thanks!

leleliu008:
Fixes #147

ebiggers:
Merged, thanks!

PR #152, opened 2022-01-02
buildsys: android apk do not support soversion

leleliu008:


ebiggers:
Why is this needed?

leleliu008:
so files included in Android apk who's name must end with .so, in other words, Android apk don't support SOVERSION.

ebiggers:
It's just a filename; it can always be renamed.  How are you getting the shared library file into your APK in the first place?  Can you elaborate on how this change is helpful for you?

leleliu008:
Android APK is Android app file. it actually is a zip file. Android developer can pack so file into it. but so filename must end with .so, beacause Android APK installer only copy file who's name end with .so to LD_LIBRARY_PATH. our library have set `-Wl,-soname=libdefalte.so.0`, if `libdefalte.so.0` linked by other library, dynamic loader would not found `libdefalte.so.0`

leleliu008:
Android apk do not support soname, Android command-line app support soname. They are not same. 
I can change this via `patchelf --set-soname libdefalte.so libdefalte.so.0`

PR #158, opened 2022-01-15
Improve handling of input overreads

ebiggers:
* deflate_decompress: error out if overread count gets too large
* deflate_decompress: rename overrun_count to overread_count
* gunzip: limit uncompressed buffer size to 1032x compressed size


ebiggers:
Merged manually

PR #163, opened 2022-02-05
fix crc32 build on arm

bkuhls:
Fixes build error

/tmp/ccAwbDTP.s: Assembler messages:
/tmp/ccAwbDTP.s:140: Error: selected processor does not support `crc32b r0,r0,r3' in ARM mode
/tmp/ccAwbDTP.s:146: Error: selected processor does not support `crc32w r3,r0,r3' in ARM mode
/tmp/ccAwbDTP.s:148: Error: selected processor does not support `crc32w r3,r3,r4' in ARM mode
/tmp/ccAwbDTP.s:150: Error: selected processor does not support `crc32w r3,r3,r4' in ARM mode
/tmp/ccAwbDTP.s:152: Error: selected processor does not support `crc32w r3,r3,r0' in ARM mode
/tmp/ccAwbDTP.s:154: Error: selected processor does not support `crc32w r3,r3,r4' in ARM mode
/tmp/ccAwbDTP.s:156: Error: selected processor does not support `crc32w r3,r3,r0' in ARM mode
/tmp/ccAwbDTP.s:158: Error: selected processor does not support `crc32w r3,r3,r4' in ARM mode
/tmp/ccAwbDTP.s:159: Error: selected processor does not support `crc32w r0,r3,r0' in ARM mode
/tmp/ccAwbDTP.s:165: Error: selected processor does not support `crc32w lr,r0,lr' in ARM mode
/tmp/ccAwbDTP.s:166: Error: selected processor does not support `crc32w r0,lr,r4' in ARM mode
/tmp/ccAwbDTP.s:170: Error: selected processor does not support `crc32b r0,r0,r3' in ARM mode
/tmp/ccAwbDTP.s:177: Error: selected processor does not support `crc32b r0,r0,r3' in ARM mode

PR #165, opened 2022-02-08
Don't use ARM CRC32 intrinsics when affected by gcc bug 104439

ebiggers:
The following build error was reported at
https://github.com/ebiggers/libdeflate/pull/163:

    /tmp/ccAwbDTP.s:140: Error: selected processor does not support crc32b r0,r0,r3' in ARM mode
    [followed by lots of similar errors]

This is caused by https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104439, a
gcc bug that has been fixed.  To work around this, disable building
crc32_arm() as dynamically-dispatched code on the affected gcc versions.

### DIFF
--- a/common/compiler_gcc.h
+++ b/common/compiler_gcc.h
@@ -122,15 +122,30 @@
 #    endif
 
      /*
-      * Determine whether CRC32 intrinsics are supported.
+      * Determine whether ARM CRC32 intrinsics are supported.
       *
-      * With gcc r274827 or later (gcc 10.1+, 9.3+, or 8.4+), or with clang,
-      * they work as expected.  (Well, not quite.  There's still a bug, but we
-      * have to work around it later when including arm_acle.h.)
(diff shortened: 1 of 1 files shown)


jmarshall:
Curious about effects on packaging here: Is this a serious enough build problem that there might be e.g. a 1.10.1 release soon to address it? Thanks for any info. (I don't have an opinion either way, my interest is just in whether 1.10 might be superseded imminently.)

ebiggers:
The relevant code has been unchanged from v1.7, so this wasn't a regression in a newer release.  It would be obvious to anyone who this affects (package won't build at all), and gcc will be fixing this properly in their next release.  So I don't consider this urgent enough to release a new version on its own.  If you need to pull in this commit earlier, please feel free to do so.

jmarshall:
Ah — I didn't realise this wasn't a new build problem. Great, that's the answer I was hoping for. Thanks!

PR #168, opened 2022-02-23
Support building with LLVM toolchain on Windows

kmilos:
Enables building in [MSYS2 Clang x86/x86_64/aarch64 environments](https://www.msys2.org/docs/environments/).

This is from the current [MSYS2 libdeflate package patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libdeflate/001-libdeflate-makefile-mingw-fix.patch) (ignore the static and shared library related differences for now).


ebiggers:
Merged, thanks!

PR #176, opened 2022-03-27
Haiku doesn't support hard links

### LINKED ISSUE #175: Haiku doesn't support hard links
Hello, today after some talks on IRC I've started working on our port for libdeflate on Haiku, I created a patch that enables a full build/install on Haiku, would be nice if someone could review it and if need be I could create a PR here to upstream our patch.

https://github.com/haikuports/haikuports/pull/6765/files

**EDIT** patch was created on comments given on IRC, so although I created it, it's a group effort :)

Begasus:
Closes #175

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,14 @@ else ifneq ($(findstring -android,$(TARGET_MACHINE)),)
    SHARED_LIB_CFLAGS  := -fPIC
    SHARED_LIB_LDFLAGS := -Wl,-soname=$(SHARED_LIB)
 
+# Compiling for Haiku
+else ifneq ($(findstring -haiku,$(TARGET_MACHINE)),)
+   SHARED_LIB         := libdeflate.so.$(SOVERSION)
+   SHARED_LIB_SYMLINK := libdeflate.so
+   SHARED_LIB_CFLAGS  := -fPIC
+   SHARED_LIB_LDFLAGS := -Wl,-soname=$(SHARED_LIB)
+   HARD_LINKS         :=
+
 # Compiling for Linux, FreeBSD, etc.
 else
    SHARED_LIB         := libdeflate.so.$(SOVERSION)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -347,8 +355,13 @@ install:all $(PKGCONFBASE)
 	install -m644 libdeflate.h $(DESTDIR)$(INCDIR)
 	install -m755 gzip$(PROG_SUFFIX) \
 		$(DESTDIR)$(BINDIR)/libdeflate-gzip$(PROG_SUFFIX)
+ifneq ($(findstring -haiku,$(TARGET_MACHINE)),)
```
ebiggers, line 358: Can you make this part conditional on `HARD_LINKS` being nonempty instead, similar to how `SHARED_LIB_SYMLINK` works just below?

Begasus: Had been thinking about that, but haven't figured that one out, thought about this as in the first part this is covered with HARD_LINKS but you didn't add some check here, could have a look for it (will get on that)

ebiggers: `HARD_LINKS` was only unset before for Windows builds, where the `install` target isn't used.  If you want the `install` target to work for this platform, making it check `HARD_LINKS` is the right thing to do.

Begasus: My idea too, this was a first "quick" solution to our problem :) busy with some other projects but if time permits I get back on it soon :)

Makefile
```diff
@@ -347,8 +355,13 @@ install:all $(PKGCONFBASE)
 	install -m644 libdeflate.h $(DESTDIR)$(INCDIR)
 	install -m755 gzip$(PROG_SUFFIX) \
 		$(DESTDIR)$(BINDIR)/libdeflate-gzip$(PROG_SUFFIX)
-	ln -f $(DESTDIR)$(BINDIR)/libdeflate-gzip$(PROG_SUFFIX)		\
-	      $(DESTDIR)$(BINDIR)/libdeflate-gunzip$(PROG_SUFFIX)
+	if [ -n "$(HARD_LINKS)" ]; then				\
```
Begasus, line 358: @ebiggers This works here and should be in the line of your previous comment, I just adjusted it to create a symbolic link instead of a hard copy.

ebiggers: Using `$(DESTDIR)$(BINDIR)/libdeflate-gzip$(PROG_SUFFIX)` for the symlink target isn't correct because `$(DESTDIR)` won't be part of the final path.  The two binaries will be in the same directory, so I think you should make it directory-relative: `ln -s libdeflate-gzip$(PROG_SUFFIX) $(DESTDIR)$(BINDIR)/libdeflate-gunzip$(PROG_SUFFIX);`

Begasus: Think this should be OK now? Thanks on the pointers!

ebiggers:
Tweaked slightly and merged in https://github.com/ebiggers/libdeflate/pull/177.  Thanks!

Begasus:
Thanks! :+1:

PR #178, opened 2022-04-15
Use stdcall on 32-bit Windows only with DLL builds

pps83:


pps83:
what was the reason to even use stdcall in the first place? I'd remove it completely

ebiggers:
> what was the reason to even use stdcall in the first place? I'd remove it completely

See commit 5f3830983c255faf72edc81746f52f9bed65c493.  libdeflate used cdecl originally, but people requested that it be changed to stdcall.

Can you elaborate on why you want it changed back to cdecl?

pps83:
> Can you elaborate on why you want it changed back to cdecl?

I find it quite strange and unusual. stdcall is the calling convention used by winapi (only on 32bit windows afaik), if you check a few most popular libs you won't see any mention of stdcall in their entire src code base. If something/somebody neesd/wants specific calling convention then they should make their own wrapper that exposes APIs as they wish imo

ebiggers:
Sure, I agree that it makes sense to not use stdcall in static library builds.  Merged, thanks!

lazka:
> Can you elaborate on why you want it changed back to cdecl?

To add to that, as a downstream this new logic is annoying as all users of the library have to set LIBDEFLATE_DLL, but none do, and the pkg-config file is also missing the macro. So we either have to patch all users+.pc file, or patch out stdcall from libdeflate and diverge from upstream.

ebiggers:
By "new logic" are you referring to 5f3830983c255faf72edc81746f52f9bed65c493 from 2019?  Or are you referring to something more recent?

ebiggers:
Ah I see; before 7e82d450694beb320e12bc8088e1f8fe7874ee31 it was only necessary to define `LIBDEFLATE_DLL` when linking to the DLL with Visual Studio, whereas now it is necessary with MinGW as well.

I think that you outvote the person who wanted VB6 support, so I'll change it back to cdecl only.

lazka:
Ah, yes, I should have been more clear re what the problem was. Thanks for the change!

PR #232, opened 2022-09-17
Fix libdeflate.pc on Windows

ebiggers:


kmilos:
Actually, the internal variables are quite useful in making the package relocatable.

Please revert the first commit, and also improve further, see e.g. [libjxl](https://github.com/libjxl/libjxl/blob/3f6ec60ce6890132c1e2ea9719de5981042a1d96/lib/jxl.cmake#L642-L663) and [libtiff](https://gitlab.com/libtiff/libtiff/-/blob/31d9e9f73dc8b6fa599c931d9ff058d8ea32795c/cmake/PkgConfig.cmake#L48-63) and their respective .pc.in templates.

Issue #236, opened 2022-10-02
CMake: Error generating build files with VS2022


rleigh-codelibre:
I'm getting this error:

```
C:\Users\rleigh\AppData\Local\JetBrains\Toolbox\apps\CLion\ch-1\222.4167.35\bin\cmake\win\bin\cmake.exe -DCMAKE_BUILD_TYPE=Debug -DCMAKE_MAKE_PROGRAM=C:/Users/rleigh/AppData/Local/JetBrains/Toolbox/apps/CLion/ch-1/222.4167.35/bin/ninja/win/ninja.exe -DCMAKE_TOOLCHAIN_FILE=C:\Users\rleigh\code\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_INSTALL_PREFIX:PATH=install -G Ninja -S C:\Users\rleigh\code\libdeflate -B C:\Users\rleigh\code\libdeflate\cmake-build-debug
-- The C compiler identification is MSVC 19.33.31630.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.33.31629/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for clock_gettime
-- Looking for clock_gettime - not found
-- Looking for futimens
-- Looking for futimens - not found
-- Looking for futimes
-- Looking for futimes - not found
-- Looking for posix_fadvise
-- Looking for posix_fadvise - not found
-- Looking for posix_madvise
-- Looking for posix_madvise - not found
-- Performing Test HAVE_STAT_NANOSECOND_PRECISION
-- Performing Test HAVE_STAT_NANOSECOND_PRECISION - Failed
-- Configuring done
-- Generating done
CMake Error:
  Running

   'C:/Users/rleigh/AppData/Local/JetBrains/Toolbox/apps/CLion/ch-1/222.4167.35/bin/ninja/win/ninja.exe' '-C' 'C:/Users/rleigh/code/libdeflate/cmake-build-debug' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:297: multiple rules generate deflate.lib [-w dupbuild=err]

  



CMake Generate step failed.  Build files cannot be regenerated correctly.
```

I'll take a closer look and open a PR once I've found what's going wrong here. /cc @bsergean

[...]

PR #237, opened 2022-10-02
Avoid lib name collision with MSVC builds

ebiggers:
When WIN32 && !MINGW, call the static library libdeflatestatic.lib so that it doesn't collide with the import library libdeflate.lib.

Fixes https://github.com/ebiggers/libdeflate/issues/236

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,8 +170,13 @@ if(LIBDEFLATE_BUILD_STATIC_LIB)
     # same way with add_subdirectory() as with other ways.
     add_library(libdeflate::libdeflate_static ALIAS libdeflate_static)
 
+    if(WIN32 AND NOT MINGW)
+        set(STATIC_LIB_NAME deflatestatic)
+    else()
+        set(STATIC_LIB_NAME deflate)
+    endif()
     set_target_properties(libdeflate_static PROPERTIES
-                          OUTPUT_NAME deflate
+                          OUTPUT_NAME ${STATIC_LIB_NAME}
                           PUBLIC_HEADER libdeflate.h)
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
I think the strategy here would work, but I would suggest checking what other libraries are doing before committing to this, because you're encoding a non-standard name for the library which would potentially impose a burden on end-users who now need to special case this (particularly those who are not using CMake and can't use the imported target).  I would strongly suggest following established convention rather than using a custom method.

Building both the static and shared libraries *at the same time*, is a strategy inherited from automake/libtool.  CMake doesn't support that by default which is why it has the BUILD_SHARED_LIBS as the means of selecting which to build.  Even with that, it doesn't get around the mutually exclusive naming of the static/import lib.

Regarding the name, it works, but `_static` might be a little nicer (but again, I would look what other libraries are doing and see what the most common conventions are).

It might be worth asking on the CMake forum for what the current recommendations are for doing this--existing practice might be a bit outdated, and the Kitware folks are usually very helpful and responsive.

Kind regards,
Roger

ebiggers:
> I think the strategy here would work, but I would suggest checking what other libraries are doing before committing to this, because you're encoding a non-standard name for the library which would potentially impose a burden on end-users who now need to special case this (particularly those who are not using CMake and can't use the imported target). I would strongly suggest following established convention rather than using a custom method.

libdeflate v1.14 and earlier already used `libdeflatestatic.lib` for the static library on Windows.  So this isn't a new name.

> Building both the static and shared libraries at the same time, is a strategy inherited from automake/libtool. CMake doesn't support that 

That is a limitation of CMake which needs to be worked around.

PR #251, opened 2022-11-24
Target architecture macro for intel/arm archs for gcc/msvc

pps83:
Co-authored-by: Shawn Hoffman <godisgovernment@gmail.com>

### DIFF
--- a/common_defs.h
+++ b/common_defs.h
@@ -45,6 +45,37 @@
 #endif
 
 /* ========================================================================== */
+/*                              Target architecture                           */
+/* ========================================================================== */
+
+#undef ARCH_X86_64
+#undef ARCH_X86_32
+#undef ARCH_ARM64
+#undef ARCH_ARM32
+
+#ifdef _MSC_VER
+#  if defined(_M_X64)
+#    define ARCH_X86_64
+#  elif defined(_M_IX86)
+#    define ARCH_X86_32
+#  elif defined(_M_ARM64)
+#    define ARCH_ARM64
+#  elif defined(_M_ARM)
+#    define ARCH_ARM32
+#  endif
+#else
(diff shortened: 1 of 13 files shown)


common_defs.h
```diff
@@ -67,6 +67,28 @@ typedef int64_t s64;
 #  endif
 #endif
 
+#if defined(_MSC_VER)
```
ebiggers, line 70: `#if defined(_MSC_VER)` => `#ifdef _MSC_VER`.

Also, can you put this this new code in its own "section" above the "Type definitions" one?  Like:

```
/* ========================================================================== */ 
/*                            Target architecture                             */ 
/* ========================================================================== */ 

(new code here)
```

Also since these are defined / undefined style macros and not 1 / 0 style macros, to be safe they should all be `#undef`ined before the correct one is `#define`d.

common_defs.h
```diff
@@ -67,6 +67,28 @@ typedef int64_t s64;
...
+#  if defined(_M_X64)
+#    define ARCH_X64
+#  elif defined(_M_IX86)
+#    define ARCH_X86
+#  elif defined(_M_ARM64)
+#    define ARCH_ARM64
+#  elif defined(_M_ARM)
+#    define ARCH_ARM
```
ebiggers, line 78: The compiler arch macros (both the MSVC ones and the gcc/clang ones) are notoriously unclear about what is 32-bit and what is 64-bit.  As long as this abstraction layer is being introduced, how about using super clear names:

* ARCH_ARM32
* ARCH_ARM64
* ARCH_X86_32
* ARCH_X86_64

pps83: addressed

ebiggers:
Merged, thanks!

PR #253, opened 2022-11-24
Reformat bitreverse_tab with print_bitreverse_tab0

pps83:
+ remove unreferenced gen_bitreverse_tab.py


ebiggers:
I don't see what the point of this is.  I think having the separate script makes more sense.  That's what I do for other generated tables / files.

pps83:
I'm going through msvc related changes, and looked into rbit related code for arm. Did this/that, and part of this reformatted the table a bit. github diff doesn't show how it differs.

image

There is no functional change, I though it was better this way. Up to you to ignore/disregard the PR.

ebiggers:
This table is for bit reversal, so IMO it makes sense for the entries to be in hex, even though it makes it a bit more verbose.

Issue #219, opened 2022-08-22
port arch-specific code to msvc


shuffle2:
when first finding this project and skimming the readme, i saw that msvc is recommended against. After looking at the code a bit I'm assuming this is mainly because cpu_features for arm + x86 is not implemented in a way that's compatible with msvc.

I'm opening this issue to see if that's the case (or, is there some other factor involved?), and maybe to remind myself to come back and implement it (not sure when I'll have time, tho).

ebiggers:
I just hardly ever use MSVC at all, so I haven't gotten the architecture-specific code working with it --- not just the runtime CPU feature detection, but all other architecture-specific code too.  Also whenever I measure performance, I'm using gcc or clang, so those are the compilers I'm implicitly optimizing for.

I don't have any plans to start using MSVC more.

ebiggers:
That's a lot of changes...  Some of them look fine.  If you open pull requests with some of the individual changes, I can consider them.  I can't merge this all at once in its current form.

It's really unfortunate that MSVC doesn't support regular C operations on vectors like gcc and clang do.  Do you know if Microsoft plans to fix that?

When adding any new restrictions to the codebase to make MSVC happy, when possible please find a way to make gcc and clang disallow the same thing too.  For example, when removing the shadowed variables, `-Wshadow` should be added to `CFLAGS`.

Why can the `__attribute__((target))` specifiers be removed? Don't they need to be replaced with something else?  How does MSVC know which CPU features to target?

Please don't do things like `#define __BMI2__`.  Just don't check for `__BMI2__` if MSVC; check for whatever it needs.

What's wrong with `const void *`?

What's wrong with `close()`, `read()`, `write()`, etc.?

ebiggers:
I'm just fixing a few of the small things myself: https://github.com/ebiggers/libdeflate/pull/222 and https://github.com/ebiggers/libdeflate/pull/221.

One question: shouldn't `__STDC_VERSION__ < 201112L` check for `defined(__STDC_VERSION__)` first?

ebiggers:
> > Why can the `__attribute__((target))` specifiers be removed? Don't they need to be replaced with something else? How does MSVC know which CPU features to target?
> 
[...]

PR #255, opened 2022-11-24
Use _target_attribute instead of __attribute__((target(t)))

pps83:


ebiggers (commented):
I guess this is okay.  Another way to do it would be make the non-empty definitions of `ATTRIBUTES` conditional on `!defined(_MSC_VER)`, or on something like `!HAVE_INTRIN_WITHOUT_TARGET`.  Your way is a bit more concise, but makes it a bit less clear what is going on (it could be unexpected that the target is ignored). Can you add a comment above `_target_attribute` that makes it super clear why an empty definition is used with MSVC?

pps83:
> Can you add a comment above `_target_attribute` that makes it super clear why an empty definition is used with MSVC?

can you pleases write, I'll just copy paste to avoid doing back-forth with edits

pps83:
added comment/description

pps83:
FYI, this is relevant to the commit: https://github.com/ebiggers/libdeflate/issues/219#issuecomment-1326386205

ebiggers:
Merged, thanks.

PR #257, opened 2022-11-25
Implement cpu_features for msvc

pps83:
+ query_arm_cpu_features from system registry
 + implement cpuid/read_xcr using intrinsics for msvc

### DIFF
--- a/lib/arm/cpu_features.c
+++ b/lib/arm/cpu_features.c
@@ -167,6 +167,58 @@ static u32 query_arm_cpu_features(void)
 	}
 	return features;
 }
+
+#elif defined(_WIN32)
+
+/*
+ * https://docs.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions
+ * "The ARM64 version of Windows presupposes that it's running on an ARMv8 or
+ * later architecture at all times. Both floating-point and NEON support are
+ * presumed to be present in hardware."
+ * Some (but not all) other features are exposed in various Windows APIs.
+ * However, the cached values of the CP15 regs can also be read out of the
+ * registry.
+ */
+
+#include <windows.h>
+#include <stdio.h>
+
+/* Note this reads cpu 0 only */
+static bool read_cp_reg(u32 reg, u64* value)
+{
+	char value_name[12];
+	DWORD value_len = sizeof(*value);
+
+	int res = snprintf(value_name, sizeof(value_name), "CP %x", reg);
+	if (res < 0 || res >= sizeof(value_name))
+		return false;
+	return RegGetValueA(HKEY_LOCAL_MACHINE,
+		"HARDWARE\\DESCRIPTION\\System\\CentralProcessor\\0", value_name,
+		RRF_RT_REG_QWORD, NULL, value, &value_len) == ERROR_SUCCESS;
+}
+
+static u32 query_arm_cpu_features(void)
+{
+	u64 reg;
+	u32 AA64ISAR0_EL1 = ARM64_SYSREG(3, 0, 0, 6, 0); // ISA Feature Register 0
+	u32 features = ARM_CPU_FEATURE_NEON;
+
+	/* Where possible, prefer supported APIs instead of looking at registry. */
+	if (IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE))
+		features |= ARM_CPU_FEATURE_PMULL;
+	if (IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE))
+		features |= ARM_CPU_FEATURE_CRC32;
+
+	if (read_cp_reg(AA64ISAR0_EL1, &reg))
+	{
+		if (((reg >> 32) & 0xf) != 0)
+			features |= ARM_CPU_FEATURE_SHA3;
+		if (((reg >> 44) & 0xf) != 0)
+			features |= ARM_CPU_FEATURE_DOTPROD;
+	}
+	return features;
+}
+
 #else
 #error "unhandled case"
 #endif
--- a/lib/x86/cpu_features.c
+++ b/lib/x86/cpu_features.c
@@ -30,6 +30,28 @@
 
 #if HAVE_DYNAMIC_X86_CPU_FEATURES
 
+#ifdef _MSC_VER
+
+static inline void
+cpuid(u32 leaf, u32 subleaf, u32 *a, u32 *b, u32 *c, u32 *d)
+{
+	int result[4];
+
+	__cpuidex(result, leaf, subleaf);
+	*a = result[0];
+	*b = result[1];
+	*c = result[2];
+	*d = result[3];
+}
+
+static inline u64
+read_xcr(u32 index)
+{
+	return _xgetbv(index);
+}
+
+#else /* _MSC_VER */
+
 /* With old GCC versions we have to manually save and restore the x86_32 PIC
  * register (ebx).  See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47602  */
 #if defined(ARCH_X86_32) && defined(__PIC__)
@@ -61,6 +83,8 @@ read_xcr(u32 index)
 
 	return ((u64)edx << 32) | eax;
 }
+
+#endif /* !_MSC_VER */
 
 #undef BIT
 #define BIT(nr)			(1UL << (nr))


lib/x86/cpu_features.c
```diff
@@ -30,6 +30,27 @@
 
 #if HAVE_DYNAMIC_X86_CPU_FEATURES
 
+#ifdef _MSC_VER
+
+static inline void
+cpuid(u32 leaf, u32 subleaf, u32 *a, u32 *b, u32 *c, u32 *d)
+{
+	u32 result[4];
```
ebiggers, line 38: `int result[4];`, so that the cast below is unnecessary.  Also leave a blank line after the declaration, please

lib/x86/cpu_features.c
```diff
@@ -62,6 +83,8 @@ read_xcr(u32 index)
 	return ((u64)edx << 32) | eax;
 }
 
+#endif
```
ebiggers, line 86: `#endif /* !_MSC_VER */`

lib/x86/cpu_features.c
```diff
@@ -30,6 +30,27 @@
...
+	*d = result[3];
+}
+
+static inline u64
+read_xcr(u32 index)
+{
+	return _xgetbv(index);
+}
+
+#else
```
ebiggers, line 52: `#else /* _MSC_VER */`

lib/x86/cpu_features.c
```diff
@@ -30,6 +30,27 @@
 
 #if HAVE_DYNAMIC_X86_CPU_FEATURES
```
ebiggers, line 31: Isn't it necessary to make `HAVE_DYNAMIC_X86_CPU_FEATURES` and `HAVE_DYNAMIC_ARM_CPU_FEATURES` be defined when `_MSC_VER`?  Otherwise this code is never built.

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+ * Some (but not all) other features are exposed in various Windows APIs.
+ * However, the cached values of the CP15 regs can also be read out of the
+ * registry.
+ */
+
+#include <Windows.h>
+#include <stdio.h>
+
+/* Note this reads cpu 0 only */
+static bool read_cp_reg(u64* value, u32 reg)
```
ebiggers, line 187: `static bool read_cp_reg(u32 reg, u64 *value)`

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+ * https://docs.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions
+ * "The ARM64 version of Windows presupposes that it's running on an ARMv8 or
+ * later architecture at all times. Both floating-point and NEON support are
+ * presumed to be present in hardware."
+ * Some (but not all) other features are exposed in various Windows APIs.
+ * However, the cached values of the CP15 regs can also be read out of the
+ * registry.
+ */
+
+#include <Windows.h>
```
ebiggers, line 183: `windows.h` (lower case), please

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+#include <stdio.h>
+
+/* Note this reads cpu 0 only */
+static bool read_cp_reg(u64* value, u32 reg)
+{
+	DWORD value_len = sizeof(*value);
+	char value_name[12];
+	if (snprintf(value_name, sizeof(value_name), "CP %x", reg) < 0)
```
ebiggers, line 191: Does this actually work?  Are you sure it doesn't need to be `CP %08x` or `CP 0x%08x`?

Also, this is the wrong way to check the return value of `snprintf()`.

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+	u32 features = ARM_CPU_FEATURE_NEON;
+
+	/* Where possible, prefer supported APIs instead of looking at registry. */
+	if (IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE))
+		features |= ARM_CPU_FEATURE_PMULL;
+	if (IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE))
+		features |= ARM_CPU_FEATURE_CRC32;
+
+	u64 reg;
+	if (read_cp_reg(&reg, AA64ISAR0_EL1))
```
ebiggers, line 210: `if (read_cp_reg(AA64ISAR0_EL1, &reg)) {`

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+	u32 AA64ISAR0_EL1 = ARM64_SYSREG(3, 0, 0, 6, 0); // ISA Feature Register 0
+	u32 features = ARM_CPU_FEATURE_NEON;
+
+	/* Where possible, prefer supported APIs instead of looking at registry. */
+	if (IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE))
+		features |= ARM_CPU_FEATURE_PMULL;
+	if (IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE))
+		features |= ARM_CPU_FEATURE_CRC32;
+
+	u64 reg;
```
ebiggers, line 209: Variable declarations at the beginning of the function, please

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+		return false;
+	return RegGetValueA(HKEY_LOCAL_MACHINE,
+		"HARDWARE\\DESCRIPTION\\System\\CentralProcessor\\0", value_name,
+		RRF_RT_REG_QWORD, NULL, value, &value_len) == ERROR_SUCCESS;
+}
+
+static u32 query_arm_cpu_features(void)
+{
+	u32 AA64ISAR0_EL1 = ARM64_SYSREG(3, 0, 0, 6, 0); // ISA Feature Register 0
+	u32 features = ARM_CPU_FEATURE_NEON;
```
ebiggers, line 201: This all assumes arm64.  I take it that Windows + arm32 is not a thing anymore?  It used to exist, apparently.

lib/arm/cpu_features.c
```diff
@@ -167,6 +167,56 @@ static u32 query_arm_cpu_features(void)
...
+ * presumed to be present in hardware."
+ * Some (but not all) other features are exposed in various Windows APIs.
+ * However, the cached values of the CP15 regs can also be read out of the
+ * registry.
+ */
+
+#include <Windows.h>
+#include <stdio.h>
+
+/* Note this reads cpu 0 only */
```
ebiggers, line 186: Is that okay?  Some ARM SoCs are weird and have different features on the different CPUs.  I know that Linux, at least, only declares support for the CPUs' common subset of features.  I'd guess that Windows does the same.  Is it possible that this method of reading the attributes of CPU 0 specifically would circumvent that and cause problems?

Maybe you want to just leave out SHA3 and DOTPROD for now, given that they aren't yet properly supported by the Windows API, apparently?

lib/x86/cpu_features.c
pps83: for now it's not built. Other chnages will enable this code.

lib/arm/cpu_features.c
pps83: snprintf check updated.

the change is based on https://github.com/shuffle2/libdeflate/commits/msvc, so I cannot tell/validate/test it

similar code from msft: https://github.com/microsoft/SymCrypt/blob/master/lib/cpuid.c

lib/arm/cpu_features.c
pps83: supposedly, it should be possible to run arm32 on arm64 cpu anyways.

lib/arm/cpu_features.c
pps83: seems like it's the way to test it: https://github.com/microsoft/SymCrypt/blob/master/lib/cpuid.c

lib/arm/cpu_features.c
ebiggers: I don't see anything that queries the registry in that code.

Anyway, I'm not going to merge something that no one has tested...

pps83: code came from https://github.com/shuffle2/libdeflate/commits/msvc
@shuffle2, did you test it on arm?

ebiggers:
I'm going to do just `lib/x86/` first, since I can actually test that.

https://github.com/ebiggers/libdeflate/pull/258 gets it all working except for the parts that would require rearranging the translation units (AVX codegen in `crc32_x86_pclmul_avx` and BMI2 codegen in `deflate_decompress_bmi2`).

pps83:
> I'm going to do just `lib/x86/` first, since I can actually test that.
> 
> #258 gets it all working except for the parts that would require rearranging the translation units (AVX codegen in `crc32_x86_pclmul_avx` and BMI2 codegen in `deflate_decompress_bmi2`).

check these changes for crc32_x86_pclmul_avx: https://github.com/ebiggers/libdeflate/compare/master...pps83:libdeflate:master-msvc might be enough to make it work.

ebiggers:
Done via 02b68139004f22f8ac67d4f72415abed61525206 and e3797751a2b96817b9242e3d4da1a5827f03b0f7, except for part that was querying the Windows registry to detect SHA3 and DOTPROD support.  Maybe that should wait until the Windows API properly supports it.

PR #266, opened 2022-11-27
msvc: use _arm_rbit intrinsic for ARCH_ARM32 to implement rbit32

pps83:
note, there is no rbit intrinsic for arm64 with ms compiler

### DIFF
--- a/common_defs.h
+++ b/common_defs.h
@@ -711,6 +711,8 @@ rbit32(u32 v)
 	return v;
 }
 #define rbit32 rbit32
+#elif defined(_MSC_VER) && defined(ARCH_ARM32)
+#define rbit32 _arm_rbit
 #endif
 
 #endif /* COMMON_DEFS_H */


common_defs.h
```diff
@@ -711,6 +711,8 @@ rbit32(u32 v)
 	return v;
 }
 #define rbit32 rbit32
+#elif defined(_MSC_VER) && defined(ARCH_ARM32)
```
ebiggers, line 714: Isn't it necessary to check whether the target CPU actually has the rbit instruction, i.e. is ARMv6T2 or later?  See what is done for gcc and clang just above this.  Or does MSVC not support targeting CPUs earlier than that?

pps83: well, rbit is a very old inst set. Not sure if you can even find devices without it. Modern windows+arm will have it for sure.

ebiggers: Well, modern Windows also wouldn't be arm32, right?

It's probably best to just leave this out if there isn't an easy way to do it correctly.  `rbit32()` isn't very important for performance anyway.

pps83: yes, I think we can close it, but it's ok to merge it. I tested with gotbolt that the asm is identical to gcc/clang asm for arm32.

ebiggers: I'm just going to close this.  If the intrinsic becomes available for arm64, we can add that.

pps83: FYI, there is `__rbitll` in arm64intr.h starting with [16.2 visual studio](https://developercommunity.visualstudio.com/t/ARM64-missing-RBIT-intrinsics/498995).

PR #273, opened 2022-12-05
Update CMakeLists.txt to fix install on Windows

brechtsanders:
Add `${CMAKE_EXECUTABLE_SUFFIX}` to support .exe extension on Windows

### DIFF
--- a/programs/CMakeLists.txt
+++ b/programs/CMakeLists.txt
@@ -47,8 +47,8 @@ if(LIBDEFLATE_BUILD_GZIP)
         set(GZIP ${CMAKE_INSTALL_FULL_BINDIR}/libdeflate-gzip)
         set(GUNZIP ${CMAKE_INSTALL_FULL_BINDIR}/libdeflate-gunzip)
         install(CODE "message(\"-- Installing: \$ENV{DESTDIR}${GUNZIP}\")")
-        install(CODE "file(CREATE_LINK \$ENV{DESTDIR}${GZIP}
-                      \$ENV{DESTDIR}${GUNZIP})" COPY_ON_ERROR)
+        install(CODE "file(CREATE_LINK \$ENV{DESTDIR}${GZIP}${CMAKE_EXECUTABLE_SUFFIX}
+                      \$ENV{DESTDIR}${GUNZIP}${CMAKE_EXECUTABLE_SUFFIX})" COPY_ON_ERROR)
     else()
         # The cmake version is too old to support file(CREATE_LINK).
         # Just compile gzip.c again to build libdeflate-gunzip.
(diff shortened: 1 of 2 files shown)


programs/CMakeLists.txt
```diff
@@ -47,8 +47,8 @@ if(LIBDEFLATE_BUILD_GZIP)
         set(GZIP ${CMAKE_INSTALL_FULL_BINDIR}/libdeflate-gzip)
         set(GUNZIP ${CMAKE_INSTALL_FULL_BINDIR}/libdeflate-gunzip)
         install(CODE "message(\"-- Installing: \$ENV{DESTDIR}${GUNZIP}\")")
-        install(CODE "file(CREATE_LINK \$ENV{DESTDIR}${GZIP}
-                      \$ENV{DESTDIR}${GUNZIP})" COPY_ON_ERROR)
+        install(CODE "file(CREATE_LINK \$ENV{DESTDIR}${GZIP}${CMAKE_EXECUTABLE_SUFFIX}
+                      \$ENV{DESTDIR}${GUNZIP}${CMAKE_EXECUTABLE_SUFFIX})" COPY_ON_ERROR)
```
lazka, line 51: I'm wondering if the `COPY_ON_ERROR` should be part of `file()`?

lazka:
this fixes the build on Windows for me, thanks

ebiggers:
The problem arises specifically with `install` and not the actual build, right?  It would be helpful if the commit subject and pull request title was something like `CMakeLists.txt: fix install on Windows`.

Also, could you update the CI configuration file `.github/workflows/ci.yml` to add the `install` step to `cross-compile-for-windows` and `windows-visualstudio-build-and-test`, so that it gets tested?

programs/CMakeLists.txt
ebiggers: Yes, I think so.  I messed this up, sorry.

programs/CMakeLists.txt
```diff
@@ -47,8 +47,8 @@ if(LIBDEFLATE_BUILD_GZIP)
         set(GZIP ${CMAKE_INSTALL_FULL_BINDIR}/libdeflate-gzip)
         set(GUNZIP ${CMAKE_INSTALL_FULL_BINDIR}/libdeflate-gunzip)
```
ebiggers, line 48: Also, please put the `${CMAKE_EXECUTABLE_SUFFIX}` here in these `set` commands, instead of embedded in the `install` command.

brechtsanders: Yes, that makes more sense

programs/CMakeLists.txt
brechtsanders: I agree, I will fix it in this PR.

ebiggers:
Done via https://github.com/ebiggers/libdeflate/pull/275.

PR #286, opened 2023-01-07
lib, programs: remove unused includes

gemesa:


programs/gzip.c
```diff
@@ -32,14 +32,11 @@
 #include "prog_util.h"
 
 #include <errno.h>
-#include <sys/types.h>
 #include <sys/stat.h>
 #ifdef _WIN32
 #  include <sys/utime.h>
 #else
-#  include <sys/time.h>
```
ebiggers, line 40: `sys/time.h` and `utime.h` are sometimes needed, see `restore_timestamps()`.

gemesa: My bad, I fixed it. I think the other headers can be safely removed.

ebiggers:
Can you please squash your two commits so that you aren't introducing a bug and then fixing it?  Thanks!

ebiggers:
Merged, thanks.

PR #312, opened 2023-05-16
Add allocate / free routines using passed in custom functions

kdt3rd:
In some environments, it may not be safe to globally change the allocator, if another thread is concurrently using libdeflate with a different allocator. As such, provide a path where the allocator and free functions can be provided with the routines instead of changing the global pointer.


meshula:
Passing the pointers is all the way through like that is concurrency safe, I like it.

ebiggers:
Hi,

I'd like to do this a bit differently.  Can you take a look at https://github.com/ebiggers/libdeflate/pull/313?  Thanks!

kdt3rd:
Commented on #313, feel free to close this one if you no longer need any thoughts from it.

ebiggers:
Closing in favor of https://github.com/ebiggers/libdeflate/pull/313

PR #313, opened 2023-05-20
Add support for per-(de)compressor malloc/free functions

ebiggers:
Allow specifying a custom memory allocator per-(de)compressor, to support use cases where there might be different users of libdeflate in the same process who want to use different memory allocators.

This is a replacement for
https://github.com/ebiggers/libdeflate/pull/312.  I changed the design slightly to be more extensible.  The allocator is specified by a new struct 'libdeflate_options' which is extensible, so the same struct will also work for special compression options/hints, which is something that has been requested in libdeflate before.  (I'm a bit hesitant to add such options, but this least makes a place to put them if it's ever done.)  I also made the pointer to the free function be stored in the (de)compressor struct, so the API for freeing is unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ option(LIBDEFLATE_GZIP_SUPPORT "Support the gzip format" ON)
 option(LIBDEFLATE_FREESTANDING
        "Build a freestanding library, i.e. a library that doesn't link to any
        libc functions like malloc(), free(), and memcpy().  Library users will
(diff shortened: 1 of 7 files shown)


kdt3rd:
This is very similar to how I did the context creation in the OpenEXR C core library to override internal functions (including malloc / free :)), so more than happy to have this path instead. I debated this path originally but then tried the simplest approach first. Thanks for considering!

kdt3rd (approved):
Thanks for considering this, will make our downstream library safer

PR #314, opened 2023-05-20
Add PHP binding to the list

dktapps:
This binding isn't complete (it only implements compression APIs, since those are easiest and most impactful for my use case), but it may be useful to others nonetheless.


ebiggers:
Merged, thanks.

It seems like this project still needs some work, though.  Besides the lack of decompression support which you mentioned, the bindings don't appear to provide a way for PHP code to allocate a compressor --- only one-shot compression is provided.  You seem to have a cache of compressors to try to make it more efficient, but it doesn't look thread-safe.

dktapps:
>  You seem to have a cache of compressors to try to make it more efficient, but it doesn't look thread-safe.

The globals used are thread-local to the worker thread doing the current request, so they'll never be accessed by more than one thread at a time.

I opted to go for this approach for the sake of userland simplicity - also, PHP object allocation adds overhead of its own.

PR #330, opened 2023-10-08
Check for availability of mmap

th-otto:
This adds a check whether mmap() is available (not all system may have it)

Only affects the test/benchmark programs

### DIFF
--- a/programs/CMakeLists.txt
+++ b/programs/CMakeLists.txt
@@ -18,6 +18,9 @@ check_c_source_compiles("#include <sys/types.h>
                          #include <sys/stat.h>
                          int main() { struct stat st; (void)st.st_atim; }"
                          HAVE_STAT_NANOSECOND_PRECISION)
+check_c_source_compiles("#include <sys/mman.h>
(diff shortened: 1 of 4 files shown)


ebiggers:
Context, please.  What platform is this for?

th-otto:
This was for Atari running FreeMiNT. But i guess there are other system involved, eg. ucLinux running on hardware without MMU, or embedded systems.

ebiggers:
Actually μClinux does have `mmap()`.  See https://www.kernel.org/doc/html/latest/admin-guide/mm/nommu-mmap.html.  So I don't think your code would actually run on μClinux.

Also in a real embedded system, most likely only the library would be used.

So, I'm not sure I want to commit to supporting the programs when `mmap()` is missing.  It seems to be a very rare use case, and I'd have no way to test it.

th-otto:
You don't have to test the programs on that system, just have to verify that the change doesn't break any already supported systems. But i agree that a missing mmap() is quite uncommon (although other packages do such checks too, maybe because they don't have fallbacks for WIN32), so feel free to close this if you are concerned about any possible regressions.

PR #334, opened 2023-11-25
CMake: Test policy existence directly

ferdnyc:
While it's possible to limit policy-change commands based on the CMake version(s) where they exist, that's not entirely future-proof. (Some very old policies may someday be removed from later releases.) It's also not necessary, as the most robust (and [recommended](https://cmake.org/cmake/help/v3.27/manual/cmake-policies.7.html#introduction)) approach is to simply test for the existence of the policy _itself_, using `if(POLICY ...)`.


ebiggers:
Merged, thanks!

PR #341, opened 2024-02-18
lib/x86/crc32: add VPCLMULQDQ implementations of CRC-32

ebiggers:
This improves CRC-32 performance on some of the latest x86 CPUs.  Three implementations are added: VPCLMULQDQ/AVX2 and VPCLMULQDQ/AVX512VL which use 256-bit vectors, and VPCLMULQDQ/AVX512F/AVX512VL which uses 512-bit vectors.  To reduce downclocking effects, the implementation with 512-bit vectors isn't used on Intel CPUs 10th generation and older.


pps83:
> To reduce downclocking effects, the implementation with 512-bit vectors isn't used on Intel CPUs 10th generation and older.

this is CPU on my laptop: 

image


it's 11th gen, but the 512-bit code gets disabled (while commit mentions 10th generation). I had to manually [comment out](https://github.com/pps83/libdeflate/commit/99744b787ef57390af1a373fb2db05d863dee6f4) the code that disables it. Otherwise, crc32 is much slower for me.

ebiggers:
> > To reduce downclocking effects, the implementation with 512-bit vectors isn't used on Intel CPUs 10th generation and older.
> 
> this is CPU on my laptop:
> 
> image
> 
> it's 11th gen, but the 512-bit code gets disabled (while commit mentions 10th generation). I had to manually [comment out](https://github.com/pps83/libdeflate/commit/99744b787ef57390af1a373fb2db05d863dee6f4) the code that disables it. Otherwise, crc32 is much slower for me.

The concern is about applications that just call the CRC functions occasionally, resulting in only occasional use of 512-bit vectors.  Older Intel CPUs do not handle that case well as they preemptively downclock and make everything slower.  You can of course run a benchmark where you CRC things continually and show that the 512-bit vectors make things faster, but that is not necessarily enough to risk the negative effects in more realistic use cases.

I've intentionally drawn the line after Tiger Lake.  The mention of 10th gen in the commit message was misleading, and I fixed the comment in the code months ago.

PR #360, opened 2024-03-18
msvc: implement rbit32 for arm32

pps83:


ebiggers (commented):
Has anything changed since https://github.com/ebiggers/libdeflate/pull/266#discussion_r1032986301?

pps83:
I forgot I made a pr already. There is no way to have windows device without rbit. There used to be some, more than 10 years ago. Now even NEON is pretty much considered to be present.

I just read other comments, I pretty much wrote what's already written there :)

pps83:
unrelated question. Regarding bsr64/bsf64 for 32-bit builds. These are going to be plain loops, while these should have been implemented in terms of 32-bit _BitScanForward/_BitScanReverse. Ok if I make a pr for that?

PR #361, opened 2024-03-19
msvc: implement optimized bsf64/bsr64 for 32-bit builds

pps83:


pps83:
I used this code to ensure that these are identical in 32-bit builds:
```
static unsigned bsr64_z(u64 v)
{
    unsigned i = 0;
    while ((v >>= 1) != 0)
        i++;
    return i;
}
static unsigned bsf64_z(u64 v)
{
    unsigned i = 0;
    for (; (v & 1) == 0; v >>= 1)
        i++;
    return i;
}

int main(int argc, char** argv)
{
    for (u64 x = 0; x < 64; ++x)
    {
        u64 z = 1ull << x;
        assert(bsr64(z) == bsr64_z(z));
        assert(bsf64(z) == bsf64_z(z));
    }
}
```

Coeur:
If you wrote test code for it, then maybe it's worth including in the project, in the `programs` folder. Like that we would see the execution of the test.

pps83:
> Thanks. If you wrote test code for it, then maybe it's worth including in the project, in the `programs` folder. Like that we would see the execution of the test.

Not sure if there is much value. All these programs are standalone executables, not some test suite. Nobody will ever try to build/execute it. The only case this code is used is 32-bit build (x86 or arm) with ms compiler. I used that code to verify that logic isn't broken running bsf/bsr for hi/lo words withing 64-bit operand.

ebiggers:
> msvc: implement optimized bsf64/bsr64 for 32-bit builds

This would be unused code, though.  These functions are only called by `bsfw()` and `bsrw()` respectively, which on 32-bit builds operate on a 32-bit word.

I probably should fold `bsf64()` and `bsr64()` into `bsfw()` and `bsrw()` to make this clear.

> Not sure if there is much value. All these programs are standalone executables, not some test suite. Nobody will ever try to build/execute it

They are part of the test suite and are wired up accordingly and run as part of the GitHub Actions CI.

But again, there's no point in this change anyway.

Issue #369, opened 2024-04-30
error: inlining failed in call to 'always_inline' in some builds


shr-project:
Not sure if this is a bug in gcc or in libdeflate, but I came across interesting side effect - build failure after -march was dropped and only -mcpu was kept with:

https://git.openembedded.org/openembedded-core/commit/?h=scarthgap&id=e64f0c1b6ac5d598a79a21de5f3060f83cb9523e

The minimal reproducer for libdeflate for me was:
```
gcc -mcpu=cortex-a76+crypto \
--sysroot=../recipe-sysroot -I. -o lib/crc32.c.o -c lib/crc32.c
```

Which fails with:
```
In file included from lib/lib_common.h:40,
                 from lib/crc32.c:171:
lib/../common_defs.h: In function 'crc32_arm_pmullx12_crc_eor3':
lib/../common_defs.h:513:1: error: inlining failed in call to 'always_inline' 'get_unaligned_le64': target specific option mismatch
  513 | get_unaligned_le64(const u8 *p)
      | ^~~~~~~~~~~~~~~~~~
In file included from lib/arm/crc32_impl.h:564,
                 from lib/crc32.c:226:
lib/arm/crc32_pmull_wide.h:199:23: note: called from here
  199 |                 crc = __crc32d(crc, get_unaligned_le64(p + 24));
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lib/../common_defs.h:513:1: error: inlining failed in call to 'always_inline' 'get_unaligned_le64': target specific option mismatch
  513 | get_unaligned_le64(const u8 *p)
      | ^~~~~~~~~~~~~~~~~~
lib/arm/crc32_pmull_wide.h:198:23: note: called from here
  198 |                 crc = __crc32d(crc, get_unaligned_le64(p + 16));
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lib/../common_defs.h:513:1: error: inlining failed in call to 'always_inline' 'get_unaligned_le64': target specific option mismatch
  513 | get_unaligned_le64(const u8 *p)
      | ^~~~~~~~~~~~~~~~~~
...
```

While restoring -march builds fine with:
```
gcc -mcpu=cortex-a76+crypto -march=armv8.2-a+crypto \
--sysroot=../recipe-sysroot -I. -o lib/crc32.c.o -c lib/crc32.c
```

[...]

PR #370, opened 2024-04-30
lib/arm: don't use explicit armv8.2-a on gcc 13.2 and later

ebiggers:
Resolves https://github.com/ebiggers/libdeflate/issues/369


shr-project:
Thanks for super fast fix, I confirm this fixed the build for me.

PR #374, opened 2024-05-08
Reuse mask for compute_x64_div_G

AtariDreams:


ebiggers:
The original code is 1 line shorter, so I think I'll just stick with that.

Issue #403, opened 2024-11-26
Heisenbug in crc code when 512-bit vectors are used with Intel C++ compiler


pps83:
I installed recently Intel's c++ compiler (I have 2024, and 2025 version). I didn't check my unit tests with intel compiler, but recently I noticed that one test fails *sometimes*. This my unit test simply calculates crc32 using zlib-ng and using libdeflate and reports timings. The crazy part is that the bug sometimes disappears: the same binary executed multiple times and sometimes libdeflate fails and sometimes returns correct crc32 value. When it fails, it always returns the same value (eg it's not always different when it fails).

Curious observation is that the build sometimes fails only when running non-optimized version. Just while writing this message I double checked different combinations and it appears that clang-cl also has the same issue!
These are the only differences that I have from public libdeflate: https://github.com/ebiggers/libdeflate/compare/master...pps83:libdeflate:master

pps83:
Here's minimal repro test: https://github.com/pps83/libdeflate/tree/crctest (you need to build with clang-cl for Debug x64).
That's the output I get *sometimes*:
```
crc32:3403043317
ERROR: libdeflate crc32:2268885453
```

ebiggers:
Thanks.  https://github.com/ebiggers/libdeflate/pull/404 should fix this.

pps83:
> Thanks. #404 should fix this.

that was quick. I'll test and report.

pps83:
fix works. FYI, added a comment here: https://github.com/ebiggers/libdeflate/pull/341#issuecomment-2504605912

ebiggers:
> I'm wondering why is that msvc build doesn't have that issue

Well, it is quite difficult for compilers to generate code that does *not* zeroize the upper bits when extending 128 bits to 512 bits, considering how the underlying instruction set works.  clang somehow managed to do so, but only at `-O0` where the generated code is very unoptimized and messy.

PR #404, opened 2024-11-27
lib/x86/crc32: fix undefined behavior in VPCLMULQDQ optimized functions

ebiggers:
The specifications for _mm256_castsi128_si256() and _mm512_castsi128_si512() are bugged, as they leave the high bits undefined instead of zeroed as would be expected.  Separate intrinsics _mm256_zextsi128_si256() and _mm512_zextsi128_si512() were later added to allow working around this defect.  Use them.

This fixes incorrect CRC checksums produced by
crc32_x86_vpclmulqdq_avx512_vl512() when built with 'clang -O0'. Other cases are not known to have been affected.

Resolves https://github.com/ebiggers/libdeflate/issues/403 Fixes: 5f2a0b4beca9 ("lib/x86/crc32: add VPCLMULQDQ implementations of CRC-32")

### DIFF
--- a/lib/x86/crc32_impl.h
+++ b/lib/x86/crc32_impl.h
@@ -78,8 +78,10 @@ static const u8 MAYBE_UNUSED shift_tab[48] = {
  *
  * gcc 8.1 and 8.2 had a similar bug where they assumed that
  * _mm256_clmulepi64_epi128() always needed AVX512.  It's fixed in gcc 8.3.
+ *
+ * _mm256_zextsi128_si256() requires gcc 10.
  */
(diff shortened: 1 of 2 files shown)


pps83:
tested and verified. Fix works, thanks

pps83 (approved):

pps83:
Can you try asan, afaik it should catch that uninitialized use. llvm+asan is broken in visual studio

ebiggers:
> Can you try asan, afaik it should catch that uninitialized use. llvm+asan is broken in visual studio

libdeflate has been tested extensively with ASAN, MSAN, UBSAN, and valgrind.  None of them can detect this case, unfortunately.

PR #416, opened 2025-02-20
Fix type mismatch in deflate_compress.c

MesuDevastator:
# In short
The type mismatch between `u32` and `unsigned int` in 
`deflate_compress.c` caused a weird compilation failure on arm-none-eabi-gcc 14.2.0 when compiling for ARM32.

# Details
```
lib/deflate_compress.c: In function 'deflate_compress_lazy_generic':
lib/deflate_compress.c:2666:49: error: passing argument 9 of 'hc_matchfinder_longest_match' from incompatible pointer type [-Wincompatible-pointer-types]
  2666 |                                                 &cur_offset);
       |                                                 ^~~~~~~~~~~
       |                                                 |
       |                                                 unsigned int *
In file included from lib/deflate_compress.c:166:
hc_matchfinder.h:191:42: note: expected 'u32 * const' {aka 'long unsigned int * const'} but argument is of type 'unsigned int *'
   191 |                              u32 * const offset_ret)
       |                              ~~~~~~~~~~~~^~~~~~~~~~
``` 
The type `u32` is somehow recognized as **`long unsigned int`** whose pointer is incompatible with `unsigned int`. Simply changing the types of the 4 variables to `u32` addresses the issue (hopefully this won't break anything).

PR #418, opened 2025-02-24
lib/deflate_compress: use u32 for match length and offsets

ebiggers:
As reported at https://github.com/ebiggers/libdeflate/pull/416, deflate_compress.c is passing a pointer to unsigned int to a function that takes a pointer to u32.  This is not compatible with unsigned int and u32 being different sizes, and was also reported to cause a compiler warning with arm-none-eabi-gcc (despite the types being the same size there).

Let's just consistently use u32 for match lengths and offsets instead of unsigned int.  This is suboptimal when sizeof(unsigned int) == 2, but that case seems to be fairly dead and libdeflate has never worked there anyway.

### DIFF
--- a/lib/deflate_compress.c
+++ b/lib/deflate_compress.c
@@ -234,7 +234,7 @@ check_buildtime_parameters(void)
 /******************************************************************************/
 
 /* Table: length slot => length slot base value */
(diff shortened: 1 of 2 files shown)


ebiggers:
Merged manually.

PR #421, opened 2025-03-09
Add support for Apple framework builds

Treata11:
To enable iOS-derived compilations, I introduced a new flag, `LIBDEFLATE_FRAMEWORK`, which is set to `FALSE` by default. I also defined the required properties for framework generation, as described in the [CMake documentation](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK).
Building is successful using the following commands:

```bash
cmake -S. -B build -DLIBDEFLATE_FRAMEWORK=ON -DLIBDEFLATE_BUILD_SHARED_LIB=OFF -DCMAKE_SYSTEM_NAME=iOS

sudo cmake --build build --target install --config Release
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,9 @@ option(LIBDEFLATE_BUILD_TESTS "Build the test programs" OFF)
 option(LIBDEFLATE_USE_SHARED_LIB
        "Link the libdeflate-gzip and test programs to the shared library instead
        of the static library" OFF)
+if(APPLE)
+    option(LIBDEFLATE_FRAMEWORK "Build as Apple Frameworks" OFF)
+endif()
 
 if(LIBDEFLATE_BUILD_TESTS)
     enable_testing()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -234,6 +237,21 @@ if(LIBDEFLATE_BUILD_STATIC_LIB)
     set_target_properties(libdeflate_static PROPERTIES
                           OUTPUT_NAME ${STATIC_LIB_NAME}
                           PUBLIC_HEADER libdeflate.h)
+    if(LIBDEFLATE_FRAMEWORK)
+        set_target_properties(libdeflate_static PROPERTIES
```
ebiggers, line 241: Is it correct for this to be specific to the static library, not the shared one?

CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ option(LIBDEFLATE_BUILD_TESTS "Build the test programs" OFF)
 option(LIBDEFLATE_USE_SHARED_LIB
        "Link the libdeflate-gzip and test programs to the shared library instead
        of the static library" OFF)
+if(APPLE)
+    option(LIBDEFLATE_FRAMEWORK "Build as Apple Frameworks" OFF)
```
ebiggers, line 48: LIBDEFLATE_APPLE_FRAMEWORK, and "Build as an Apple Framework"

CMakeLists.txt
Treata11: Dynamic libs aren't allowed on iOS.

ebiggers:
Merged, thanks.

PR #424, opened 2025-04-21
Framework Resources

Treata11:
In the previous PR-#421, I didn't see that Deflate comes with a `libdeflate-config.cmake` script instead of a `find-deflate.cmake` ... 
So, to make the build process of downstream projects for Apple platforms more convenient, I added it to the Framework bundle (alongside the Readme & the License)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,32 @@ if(LIBDEFLATE_FREESTANDING)
     add_definitions(-DFREESTANDING)
 endif()
 
+function(configure_framework libtype)
+    if(LIBDEFLATE_APPLE_FRAMEWORK)
+        set_target_properties(${libtype} PROPERTIES
+            FRAMEWORK TRUE
+            FRAMEWORK_VERSION "${VERSION_STRING}"
+            PRODUCT_BUNDLE_IDENTIFIER "github.com/ebiggers/libdeflate/${libtype}"
+            XCODE_ATTRIBUTE_INSTALL_PATH "@rpath"
+            XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY ""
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -271,6 +291,7 @@ if(LIBDEFLATE_BUILD_SHARED_LIB)
                           PUBLIC_HEADER libdeflate.h
                           C_VISIBILITY_PRESET hidden
                           SOVERSION 0)
+    configure_framework(libdeflate_shared)
```
Treata11, line 294: In the [previous response](https://github.com/ebiggers/libdeflate/pull/421#r1987659390), I oversaw the fact that one can build frameworks for macOS, so setting up the properties for 
shared-libs do make sense...

CMakeLists.txt
```diff
@@ -221,6 +246,15 @@ set(LIB_INCLUDE_DIRS
...
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_FULL_INCLUDEDIR}>)
 
+# Define resource files for Apple Framework
+set(LIBDEFLATE_RESOURCES
+    "${CMAKE_CURRENT_SOURCE_DIR}/README.md"
+    "${CMAKE_CURRENT_SOURCE_DIR}/COPYING"
+    "${CMAKE_CURRENT_BINARY_DIR}/libdeflate-config.cmake"
+    "${CMAKE_CURRENT_BINARY_DIR}/libdeflate-targets.cmake"
+)
+message(STATUS "Resource files: ${LIBDEFLATE_RESOURCES}")
```
ebiggers, line 256: Can you make this message conditional on LIBDEFLATE_APPLE_FRAMEWORK so that it doesn't show up for everyone else?

PR #425, opened 2025-04-24
Avoid implicit casts

pps83:
With the change the code can be compiled with c++ compiler for correctness checks


ebiggers:
Well, it's C code.  Not C++.

pps83:
> Well, it's C code. Not C++.

compiling with c++ compiler often can shows code issues. brotli, for instance, compiles cleanly with a c++ compiler

ebiggers:
Not going to proceed with this for now.  It can be reconsidered if there's real demand for building libdeflate with a C++ compiler (in which case it would need to be tested in CI).

Issue #394, opened 2024-10-23
Compile error when building via zig build


lambdadog:
I'm not entirely sure if this is an issue for upstream, but since libdeflate advertises that it's able to be built via just copying the source files into your project, I was hoping you might able to advise/determine whether an upstream fix is needed.

The primary relevant error seems to be `always_inline function '<compiler intrinsic>' requires target feature 'evex512', but would be inlined into function '<function>' that is compiled without support for 'evex512'` (see error log attached below for specific examples) and indeed, if the project is built with `zig build -Dcpu=baseline+evex512` it compiles without issue. See [omen-mc/libdeflate](https://github.com/omen-mc/libdeflate) for repro (packacking of libdeflate via zig build system).

I'm not an expert when it comes to working with cpu-specific features, but it seems to me that perhaps "evex512" needs to be added to some `_target_attribute` lists (such as [here](https://github.com/ebiggers/libdeflate/blob/2335c047e91cac6fd04cb0fd2769380395149f15/lib/x86/adler32_impl.h#L98)? I'm not exactly sure why building with clang via `zig build` would cause this issue to show up when libdeflate (aiui) builds with clang normally without issues, however, which is making me not particularly confident in my answer.

<details>
<summary>Error log</summary>

```
/home/lambdadog/.cache/zig/p/12203648ce8079997389a2bd39f847ae9dad666594173d499a061e813234c4c8a942/lib/x86/adler32_template.h:197:21: error: always_inline function '_mm512_set1_epi8' requires target feature 'evex512', but would be inlined into function 'adler32_x86_avx512_vl512_vnni' that is compiled without support for 'evex512'
 const vec_t ones = VSET1_8(1);
                    ^
/home/lambdadog/.cache/zig/p/12203648ce8079997389a2bd39f847ae9dad666594173d499a061e813234c4c8a942/lib/adler32.c:128:12: note: in file included from /home/lambdadog/.cache/zig/p/12203648ce8079997389a2bd39f847ae9dad666594173d499a061e813234c4c8a942/lib/adler32.c:128:
#  include "x86/adler32_impl.h"
           ^
[...]

PR #427, opened 2025-05-20
lib/x86/cpu_features.h: drop evex512 on gcc-16

trofi:
gcc-16 `master` branch dropped support for `evex512` as:
    https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=c052a6f4a1c803cb92147ff98fb91cf3511e0856

As a result libdeflate build started failing as:

    In file included from /build/source/lib/adler32.c:28:
    /build/source/lib/x86/adler32_template.h:135:12: error: attribute 'target' argument 'no-evex512' is unknown
      135 | ADD_SUFFIX(reduce_to_32bits)(vec_t v_s1, vec_t v_s2, u32 *s1_p, u32 *s2_p)
          |            ^~~~~~~~~~~~~~~~

The change adds upper build to `evex512` for `gcc`.

### DIFF
--- a/lib/x86/adler32_impl.h
+++ b/lib/x86/adler32_impl.h
@@ -82,7 +82,7 @@
  */
 #  define adler32_x86_avx512_vl256_vnni	adler32_x86_avx512_vl256_vnni
 #  define SUFFIX				   _avx512_vl256_vnni
-#  define ATTRIBUTES		_target_attribute("avx512bw,avx512vl,avx512vnni" NO_EVEX512)
+#  define ATTRIBUTES		_target_attribute("avx512bw,avx512vl,avx512vnni")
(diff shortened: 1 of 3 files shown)


pinskia:
I think evex512 will be removed from GCC 15.2.0 also. So it might just make sense to remove all references to it since it will never be supported in HW alone.

trofi:
Sounds good! Added extra `lib/x86: drop any mention of evex512` commit to completely remove it.

ebiggers:
I'll merge this.  Unfortunately this will regress https://github.com/ebiggers/libdeflate/issues/394 (FYI @lambdadog), but they are just going to have to stop using `-mno-evex512` anyway.

